PR #55, opened 2014-12-16
MAKE_FUNCTION checks for Qualified Name

mancoast:


mancoast:
Extra period in test file names.  Requires resubmit

Issue #39, opened 2013-09-27
Unsupported opcode: LOAD_BUILD_CLASS


schlamar:
Cannot decompile for Python 3.3. It fails with the above message.

ReDucTor:
This code should really be wrapped in an if() for the specific version, it adds confusion when trying to work out what CALL_FUNCTION is doing.

PR #59, opened 2015-04-20
Add support for LOAD_BUILD_CLASS op-code

mancoast:
Issue #39 

This approach initially shows good results, but some review/input is needed.


Azeirah:
On running pycdc on a python 3 file, this doesn't perform any differently. I get an `Unsupported opcode: LOAD_BUILD_CLASS` error still

mancoast:
please provide sample code for the STORE_LOCALS for fix.

Azeirah:
Oh, I've just made an empty case statement like following.
```
...
case STORE_LOCALS: 
    break;
...
```

This generated code was good enough to fix by hand, so I didn't bother to implement a proper solution.

PR #63, opened 2015-09-05
add STORE_LOCALS support

zjjott:


vstoykov:
+1

EhsanKia:
This doesn't actually work? When I add this, I get 

```
Warning: block stack is not empty!
Warning: Stack history is not empty!
Segmentation fault (core dumped)
```

PR #75, opened 2016-01-19
The big pull request

Djamana:
Hmm maybe one time I'll maybe able to get create a little bit more specific pull but at the moment I'm happy to get it finally done 'somehow'.
It's a little pitty that I can't select what to include and what not.
However before fiddling around I better just do it.


zrax:
Wow, I admit haven't looked at all of this yet, but here are some comments about what jumped out at me right away for starters:

1. Please don't commit generated files.  They are generated at build-time (or configure time) rather than being checked in so they don't have to be kept up-to-date separately from the files that generate them.  This includes the compiled executables, the map_*.cpp files, and the MSVC project files (which don't even appear to be CMake generated?)
2. The extra Windows .cmd scripts are unnecessary -- CMake handles the generation of these files automatically when it generates your project files / makefiles.
3. Your editor seems to be configured to use tabs instead of spaces -- please ensure you commit with 4-space indent to be consistent with the rest of the code base.
4. This is a cross-platform application -- please don't use Windows-specific functions and headers without appropriate reason and conditional compilation as necessary.
5. It's better to separate different sets of unrelated work into different PRs.  For example, behavior changes should be in a separate PR from cleanup or refactoring changes.  At the very least, they should be in separate commits.
6. I don't quite understand the purpose of the log functions...  What do they provide that fprintf by itself does not?  Why is the function that writes the decompiled output also called "log"?
7. Using an enumeration instead of normal C++ exception classes is highly unusual in modern code.  Why not instead use a single exception class that has the enum as a data member?

Djamana:
>1.
Oh just installed and tried CMake right now. Wow cool works like a charm.
Hmm why haven't I have done that earlier? ;)
Okay I agree - seeing CMake working like that - all that clumsy MSVC project files I create are just like clutter around. I'll leave them out.

The only point I don't go along with is about  *publishing the *.exe*. 
If I just want to use it, I love to get compiled executables and be done. Even if it's a little outdate/ buggy or what ever. If I want to do things quickly or try out something the compiled executables are gold.
There was only this 'Easy Python Decompiler' that I got to run in first place. All the other solution were kind of 'blocked' since there was no exe or *.py you can  just download and run it.
Getting the source and compile it is nice for extending and debugging but for just the first contact its not very welcoming to me if build-ya-own is the only option.
(Especially on this monolithic window were an exe is kind'a universal runnable. )

>2. Hmm so far  haven't figured out or seen when CMake runs the 'Test'-Script
>3. Oops take care of that.
However just set Options/Editor/C++/Tabstop to 'insert space' and Tab= 4 Spaces that'll hopefull fix this.
>4. Well haven't try to compile it on some other system. Of course would be awesome to keep it  cross-platform. Hmm fixed 2 build warnings about 'unsecure' function but maybe this '_s'-crap breaks portability. Also added #include <share.h> and made fopen to share_all since there is no need zu block the file as long as it is open. Of course I would welcome ya point to out what of the changes is troublesome and are is windows-specific crap.
>5. Well who ya telling, I'm also very unsatisfied by big PR. However did saw any way to opt-out stuff like for ex the map_*.cpp from PR or very how to delete stuff from commits. Probably a nice thing that github however need probably some practice and experience to use it more fine grained.
To me its often a matter of focus. I enjoy more to do one thing a one time. So ex. when coding I not so pleasured to spend hours on learning setting up github  / interrogating it into Visual Studio  + fiddling out how it works even when I see it's very rewarding.
However the vision is there, how i'll be optimal. :dancer: 

>6. Well glad ya asking. It's about organization. So far they are not providing any additional functionally to printf (When you make them inline them might as well not really 'visible' in the compiled/resulting code) . However they create potential. Even if at the current time all the log() logErr() logVerb() are pointing to fprintf(), now it's really no big deal anymore to do changes like maybe put "ERROR: " in front of all logErr output to make it shown in red or not show at all.
On the other hand this improves readably and clearness of the code. Focusing more on the intent and on what is done instead of how (... and to push that how to the max - for ex. what the asm code might look like.) Code get's more speaking and language like this. Also the other macros/templates I added heading this way. Hiding repetitive implementation clutter while keeping in important string / function name so the purpose of the code to you or some else look it gets clear more quickly.
Beside hiding and encapsulating stuff that belongs together should stay together. =>'Locality'
So i keep the 'implementation clutter' as close as well as central to were it's used.
Beside of course functions and templates - macro's are a great addition to make the code more speaking, more to the point and more beautiful.

Well still learning/exercising the good way of using them, as well as not using them. ;)

>7.
>Using an enumeration instead of normal C++ exception classes is highly unusual in modern code.
>Why not instead use a single exception class that has the enum as a data member?

Sounds good. So put that into a class. Idea behind was to 
1. add at least one global exception handler on the top of the 'call stack' so nullpointer or access violations are caught there and
2. to be able to exit a function or the whole program in case of an unrecoverable error via throw while 
3. passing a specific exit code.

Don't need to pass along additional information like the for ErrorText, ErrLocation...
For that the enum was sufficient. Keeping the error codes nicely group together via the enum type.
Logerr('Error message') is shown right before the throw and the ERR_CODE_Constant_Name is used for self-documentation.  Static class constant does the same, but then I always have to write the Classname::ERR_code. Hmm wor

ReDucTor:
Just to progress this
* 63bd337 (reject - generated files)
* a6c700b (reject - generated files)
* 12cf6b5 (reject #2 mentioned in comments, also common git installs for windows include, bash that runs scripts, pull request #78 has tests that work on windows)
* c5977f6 (reject - no real value - issue #79 adds reasonable debate for improvement, and this hides printf format issues with variadic templates)

Overall: reject all

Follow up from comments left by @Djamana

1. Publishing binaries in SCM (Source control management) gets big fast, you should expect alot of commits to be established for one release. Binaries should be associated with a 'tag', github provides this functionality, and it should be used. We should not bloat the source code repo to support this, it adds zero-benefit
2. CMake usage for testing needs improving, however I suggest the usage of bash on windows (git often comes with an installation of this); the batch file adds extra maintenance overhead which for the scale of this project is not good
3. No comment
4. CRT in-secure warnings while semi-useful are not very portable for MSVC and identify alot of false positives, instead proper static analysis should be used to resolve this problems rather then compromising and complicating cross-platform code. (Nearly every cross-platform code base I've dealt with commercial or open-source disables these CRT warnings.)
5. No comment; github is painful; git is not beginner (or expert at times) friendly
6. Nice motivation, other changes blended doesn't put strong motivation, issue #79 should resolve some of this separation.
7. Global 'generic' catches can make it harder to catch issues, part of the reasoning behind only catch what you know is don't handle the 'unexpected', handle what you can, less bugs easier to reason about error handling code (which is always painful). As for specific 'error code', windows, linux, etc will all on a segfault/crash cause an invalid error code to be generated, trying to do the same thing is unnecessary and more code to maintain.

While github makes life painful for incremental pull requests, I'd suggest using branches for 'big' pull requests.

PR #78, opened 2016-05-16
Updating testing and (minor) MSVC improvement

ReDucTor:
I have a bunch of fixes and improvements coming for future pushes, doing an overhaul of the way tests are run so then it can be verified what changed.

TODO: (Minor) - Let me know if any required for pull-request
* Cover more python versions (Not covered before)
* Fully verify non-stolen decompyle stolen tests are covered (If pull accepted, will bug)
* Documentation on how to update baseline (copy-based)


mancoast:
Regarding test cases from distributions 1.5.2 - 3.4.2:
I have been using the Xeon Phi to execute many tests in parallel.

Some preliminary sorting:

These tests produced output, [**not** validated]
https://github.com/mancoast/CPythonPyc_test/tree/master/cpython

These tests crashed [validated as **incorrect**]:
https://github.com/mancoast/CPythonPyc_test/tree/master/crash

These tests produced bad output [validated as **incorrect**]:
https://github.com/mancoast/CPythonPyc_test/tree/master/fail

ReDucTor:
Thanks for wider test suite. I welcome and wish to integrate this larger suite however should it be a current CL blocker or new commit?

mancoast:
those files were dumped from the full regression test foreach python distribution.
what do you mean by current CL blocker or new commit?

ReDucTor:
What I mean by blocker or new commit is should your suggestion block committing/merge the change I currently have in the pull request (to make it more ideal with CPythonPyc_test regression tests), or should that be done as a later pull request/commit

mancoast:
I like your suggestions with the test suite. the additional test cases from cpythonpyc-test do not block this commit. The additional test cases require some restructuring.

rocky:
I just came across this. It is a pity this hasn't been acted upon. If someone wants to rework them in the context of uncompyle6's test framework, I'd support that. 

I see that a number of these come from decompyle's test suite, for example `test_applyEquiv.py`.  This particular test exists under in `python-uncompyle6/test/simple_source/bug22/02_apply_equiv.py`. 

However the better way used in testing now would be to adapt this program to add asserts, so that when this program runs, it checks itself. For example, instead of: `no_apply(*args, **args)` we would have `no_apply(*args, **args) == ...`.

PR #81, opened 2016-07-10
Add comparison argument to disassembly output

Saruta:
Output format:
`9       COMPARE_OP              2: '=='`

### DIFF
--- a/bytecode.cpp
+++ b/bytecode.cpp
@@ -167,6 +167,10 @@ bool Pyc::IsJumpOffsetArg(int opcode)
            (opcode == Pyc::FOR_LOOP_A);
 }
 
+bool Pyc::IsCompareArg(int opcode)
+{
+    return (opcode == Pyc::COMPARE_OP_A);
+}
 
 void print_const(PycRef<PycObject> obj, PycModule* mod)
 {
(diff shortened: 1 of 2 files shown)


bytecode.cpp
```diff
@@ -349,7 +357,9 @@ void bc_disasm(PycRef<PycCode> code, PycModule* mod, int indent)
                 print_const(code->getCellVar(operand), mod);
             } else if (Pyc::IsJumpOffsetArg(opcode)) {
                 fprintf(pyc_output, "%d (to %d)", operand, pos+operand);
-            } else {
+            } else if (Pyc::IsCompareArg(opcode)) {
+		fprintf(pyc_output, "%d: '%s'", operand, cmp_strings[operand]);
```
zrax: Please add a range check to avoid accessing `cmp_strings` out of bounds

zrax:
Please use only spaces for indentation (no tabs) to be consistent with the rest of the code base.

Saruta:
Changed, sorry for that.

bytecode.cpp
```diff
@@ -349,6 +357,8 @@ void bc_disasm(PycRef<PycCode> code, PycModule* mod, int indent)
                 print_const(code->getCellVar(operand), mod);
             } else if (Pyc::IsJumpOffsetArg(opcode)) {
                 fprintf(pyc_output, "%d (to %d)", operand, pos+operand);
+            } else if (Pyc::IsCompareArg(opcode)) {
+                fprintf(pyc_output, "%d: '%s'", operand, cmp_strings[operand]);
```
zrax: Please add a range check to avoid accessing `cmp_strings` out of bounds

PR #88, opened 2016-08-14
add "await" (for 3.5 when @asyncio.coroutine Decorator is not used), …

AraHaan:
…and yield from for when it is.

At least now lets hope this thing gets support for decorators on top of functions (for coroutines mainly) without wrapping them first.


zrax:
The ASTKeyword class is for standalone keyword nodes that have no parameters.  Both `await` and `yield from` will need to be handled in separate node types (see for example how `yield` is handled).

PR #170, opened 2020-02-10
Fixed a crash when a Stringref was an empty string (length 0)

Almamu:
Signed-off-by: Alexis Maiquez Murcia <almamu@almamu.com>


Almamu:
By the way, this is a regression after the change
f88869fb1729d88db368ed7a6b12b870aca56dad

Issue #35, opened 2013-08-07
another bug : program abends


daishaofei:
when do the file program abends,file like this on the web : https://www.dropbox.com/s/tn2krl17h4v7134/Root.pyc

PR #171, opened 2020-02-10
Fixed crash on stack_history being empty

Almamu:
Signed-off-by: Alexis Maiquez Murcia <almamu@almamu.com>

Should fix issue #35 
I've added a warning message because I'm not exactly sure that this situation is expected and/or desirable.


Almamu:
Yeah, crashing is not the best option, I do agree that it is a dirty hack. For now my objective is to get the python 2.5 parser improved to properly decompile some files, so I'll try to improve error handling for this kind of situations too so we can properly report those errors to prevent these kind of crashes (and cleanup all the ifs to check for emptyness). If you don't mind, I'll keep sending push requests with individual fixes so you can review them when you have the time.

PR #175, opened 2020-10-16
Handle NaN and infinity values

dotjrich:
Right now, NaN/infinity values will produce "nan" and "inf", but
Python doesn't allow those in source code. This change will wrap
those values in float(''), which is allowed.

Tests for Python 2.7 and 3.8 have been added as well.


dotjrich:
I was relying on just wrapping the parsed value in `float('')`, but it looks like MSVC produces `-nan(ind)` in the NaN case instead of `-nan`. I'll rework how I'm doing things and directly output nan and inf with the appropriate sign.

zrax (approved):
Looks good, thanks for the contribution!

PR #178, opened 2020-10-18
Fixes possible segfault scenario when processing JUMP_FORWARD op

### LINKED ISSUE #119: if 1 else
Apparently Python produces different code for `... if 1 else ...`

decompiling: 
```1 if 1 else __file__```

gives: 

```pycdc test/bytecode_2.7/03_if_true_else.pyc
# File: 03_if_true_else.pyc (Python 2.7)
Segmentation fault (core dumped)```

dotjrich:
The current code was always pulling the top of the blocks stack.
In certain cases - like when in BLK_MAIN - the stack may be empty.

Tests for Python 2.7 and 3.8 have also been added.

Fixes #119


dotjrich:
I just realized there is a little more work to be done before this behaves properly. I'll work on that and open a new PR when it's ready.

Deledrius:
FYI for future reference: you can push additional changes to the same branch for the PR and it'll be updated with no need to open a new one. :)

PR #182, opened 2020-10-22
Adds support for SETUP_ANNOTATION op and variable annotations

### LINKED ISSUE #169: Python 3.6+ SETUP_ANNOTATION
Python 3.6+ `SETUP_ANNOTATION` is not handled. Here is code that will create such an opcode:

```pythion
x: int
```

dotjrich:
Tests have also been added.

Fixes #169


ASTree.cpp
```diff
@@ -2117,10 +2118,36 @@ PycRef<ASTNode> BuildFromCode(PycRef<PycCode> code, PycModule* mod)
...
+                    // Python handles a varaible annotation by setting:
+                    // __annotations__['var-name'] = type
+                    bool found_annotated_var = false;
+                    if (variable_annotations) {
+                        if (dest->type() == ASTNode::Type::NODE_NAME) {
+                            if (dest.cast<ASTName>()->name()->isEqual("__annotations__")) {
+                                found_annotated_var = true;
+                            }
+                        }
+                    }
```
zrax, line 2131: I would probably put this in a single expression, rather than three nested `if` blocks:

```c++
    const bool found_annotated_var = (variable_annotations && dest->type() == ASTNode::Type::NODE_NAME
                                      && dest.cast<ASTName>()->name()->isEqual("__annotations__"));
```

ASTree.cpp
```diff
@@ -2117,10 +2118,36 @@ PycRef<ASTNode> BuildFromCode(PycRef<PycCode> code, PycModule* mod)
...
+
+                    if (found_annotated_var) {
+                        // Annotations can be done alone or as part of an assignment.
+                        // In the case of an assignment, we'll see a NODE_STORE on the stack.
+                        if (curblock->nodes().back()->type() == ASTNode::Type::NODE_STORE) {
```
zrax, line 2136: Need to check if nodes is !empty first?

ASTree.cpp
dotjrich: Done! This looks much cleaner now :)

ASTree.cpp
dotjrich: Good catch; done!

zrax (approved):

Issue #185, opened 2020-10-23
Chained assignments


Aralox:
```
a = b = sin(x) + 3*sideEffect()
```
decompiles to
```
a = sin(x) + 3*sideEffect()
b = sin(x) + 3*sideEffect()
```

I've got a fix ready for this, will PR after #184 is done.

PR #186, opened 2020-10-23
Support chained assignment statements, e.g. `a = b = c`

Aralox:
We know when we have begun a chained assignment when we process a DUP_TOP with non-null on the stack. Push a NODE_CHAINSTORE onto the stack when this happens, and keep it 'floating' on top of the stack for all STORE_X operations until the stack is empty.
To support versions of Python <= 2.5 which use DUP_TOP in more places, I modified ROT_TWO, ROT_THREE and ROT_FOUR to get rid of NODE_CHAINSTORE on the stack if it is present.
Issue #185 

[Opened this a little early on top of the other branch]


ASTree.cpp
```diff
@@ -679,7 +684,22 @@ PycRef<ASTNode> BuildFromCode(PycRef<PycCode> code, PycModule* mod)
             }
             break;
         case Pyc::DUP_TOP:
-            stack.push(stack.top());
+            {
+                if (stack.top().type() != PycObject::TYPE_NULL) {
```
zrax, line 688: This should be `ASTNode::NODE_INVALID`.

Also, this could be flattened to not require nested `if`s

ASTree.cpp
```diff
@@ -2243,10 +2309,27 @@ PycRef<ASTNode> BuildFromCode(PycRef<PycCode> code, PycModule* mod)
...
+                        curblock.cast<ASTIterBlock>()->setIndex(tup);
+                    } else if (stack.top().type() == ASTNode::NODE_CHAINSTORE) {
+                        auto chainStore = stack.top();
+                        stack.pop();
+                        append_to_chain_store(chainStore, tup, stack, curblock);
+                    } else {
+                        curblock->append(new ASTStore(stack.top(), tup));
+                        stack.pop();
+                    }
+                }
```
zrax, line 2332: Please rebase to remove the parts of #184 in this PR.

ASTree.cpp
```diff
@@ -2306,6 +2389,17 @@ PycRef<ASTNode> BuildFromCode(PycRef<PycCode> code, PycModule* mod)
     return new ASTNodeList(defblock->nodes());
 }
 
+static void append_to_chain_store(PycRef<ASTNode> chainStore, PycRef<ASTNode> item, FastStack& stack, PycRef<ASTBlock> curblock)
```
zrax, line 2392: Please wrap long lines to around 80-100 columns.

ASTree.cpp
```diff
@@ -2306,6 +2389,17 @@ PycRef<ASTNode> BuildFromCode(PycRef<PycCode> code, PycModule* mod)
...
+{
+    stack.pop();    // ignore identical source object.
+    chainStore.cast<ASTChainStore>()->append(item);
+    if (stack.top().type() == PycObject::TYPE_NULL) {
```
zrax, line 2396: Should be `ASTNode::NODE_INVALID`.

Aralox: I'm not sure how to get that to work - TYPE_NULL is 48 (chr('0')) which is what stack.top() is. E.g. in my debugger
image
How do I read that as NODE_INVALID instead?

zrax: Whoops, you're right.  Sorry for the confusion

Aralox: no worries

Aralox:
Thanks for your code reviews Michael, I've had a lot of fun working on pycdc the last few weeks for hacktoberfest :)

zrax (approved):

zrax:
Thanks again for the contribution!

PR #202, opened 2021-10-09
added python 3.10 support

ghost:
added "temporary support" for python 3.10, however some of the newer opcodes are not in `bytes/python_310.map` to prevent conflict with the undeclared pycdc opcodes , however im working on them, any suggestions on how i should add them to ASTree would be appreciated thanks!


zrax:
You can add the new opcodes to `bytecode_ops.inl`.  The order doesn't matter much, except that the opcodes with no arg (`OPCODE()`) need to be before those with an arg (`OPCODE_A()`).  I usually just add new opcodes to the end of these two lists for simplicity.

GawdOfROFL:
PTAL

PR #204, opened 2021-10-10
Adds container support to pymultic

dotjrich:
Using the '-c' argument will - if possible - fetch and use a
container for each version of Python specified in the arguments
list.

### DIFF
--- a/scripts/pymultic
+++ b/scripts/pymultic
@@ -54,7 +54,10 @@ PYVER_CONTAINERS = {
     '3.10',
 }
 CONTAINER_EXES = ['podman', 'docker']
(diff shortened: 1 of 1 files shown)


dotjrich:
> An additional note: I noticed this creates the output .pyc file as root (presumably because the container runs as root)... This should probably be fixed to avoid odd permission issues.

Hmm, this hasn't actually been something I've encountered. I suspect that it might be due to my use of podman instead of docker. Which are you running?

Anywho, let me play around with this. We should be able to chown it somehow.

zrax:
Yep, I tried with docker, as I'm unfamiliar with podman.

PR #209, opened 2021-10-24
Adds support for building our own Python containers

dotjrich:
This will be triggered when `-c` is used with pymultic and there is
no official Python container available for the version(s) specified.

Tested on:
* Fedora 34 + podman
* Debian 10 + docker

Tested with Python versions:
* 1.0.1
* 1.1
* 1.2
* 1.3
* 1.4
* 1.5.2
* 1.6.1
* 2.0.1
* 2.1.3
* 2.2.3
* 2.3.7
* 2.4.6
* 2.5.6
* 2.6.9
* 3.0.1
* 3.1.5

### DIFF
--- a/scripts/pymultic
+++ b/scripts/pymultic
@@ -60,6 +60,20 @@ CONTAINER_EXE_EXTRA_ARGS = {
 }
 
 
+def get_container_exe():
+    container_exe = None
+    for ce in CONTAINER_EXES:
+        if shutil.which(ce) is not None:
+            container_exe = ce
+            break
+
+    if container_exe is None:
+        print('Cannot find {} in $PATH'.format(' or '.join(CONTAINER_EXES)))
+        return sys.exit(1)
+
+    return container_exe
+
+
 def fetch_python(snekdir, version):
     realver = PYVERS[version]
     if version in ('1.0', '1.1'):
@@ -142,6 +156,35 @@ def acquire_python(snekdir, version):
     return pyexe
 
 
+def build_python_container(snekdir, version):
+    realver = PYVERS[version]
+    snek = 'Python-{}'.format(realver)
+    builddir = os.path.join(snekdir, snek)
+    container_exe = get_container_exe()
+    py_container_tag = 'python:{}'.format(realver)
+
+    if subprocess.call([container_exe, 'image', 'inspect', py_container_tag],
+                       stdout=subprocess.DEVNULL,
+                       stderr=subprocess.DEVNULL) == 0:
+        return
+
+    fetch_python(snekdir, version)
+
+    print('Building Python {} container...'.format(realver))
+    logfile = os.path.join(snekdir, '{}.build.log'.format(snek))
+    install_target = 'fullinstall' if version == '3.0' else 'install'
+    with open(logfile, 'wb') as log:
+        if subprocess.call([container_exe, 'build',
+                            '--build-arg', 'python_version={}'.format(realver),
+                            '--build-arg', 'install_target={}'.format(install_target),
+                            '-f', os.path.join(snekdir, 'Dockerfile.pybuild'),
+                            '-t', py_container_tag, snekdir], stdout=log, stderr=log) != 0:
+            print('...Container build failed.  See {} for details'.format(logfile))
+            sys.exit(1)
+
+    shutil.rmtree(builddir)
+
+
 def local_compile(snekdir, ver, infile):
     pyexe = acquire_python(snekdir, ver)
     proc = subprocess.Popen([pyexe, '-c', 'import sys; print(sys.version)'],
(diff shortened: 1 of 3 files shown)


scripts/pymultic
```diff
@@ -142,6 +156,35 @@ def acquire_python(snekdir, version):
...
+    if subprocess.call([container_exe, 'image', 'inspect', py_container_tag],
+                       stdout=subprocess.DEVNULL,
+                       stderr=subprocess.DEVNULL) == 0:
+        return
+
+    fetch_python(snekdir, version)
+
+    print('Building Python {} container...'.format(realver))
+    logfile = os.path.join(snekdir, '{}.build.log'.format(snek))
+    install_target = 'fullinstall' if version == '3.0' else 'install'
```
zrax, line 175: Interesting...  Why the difference here for 3.0?

scripts/pymultic
```diff
@@ -223,14 +255,38 @@ def container_compile(ver, infile):
...
-    proc.communicate()
+    if ver in {'1.0', '1.1', '1.2', '1.3', '1.4'}:
+        # The hard way -- hope your code is safe...
+        srcdir = os.path.dirname(os.path.realpath(infile))
+        comptmp = os.path.join(indir, 'pymc_temp.py')
+        if os.path.exists(comptmp):
+            os.unlink(comptmp)
+        shutil.copyfile(infile, comptmp)
+        proc = subprocess.Popen([container_exe, 'run'] + CONTAINER_EXE_EXTRA_ARGS[container_exe] +
+                                ['--rm', '--name', '{}'.format(outfile),
```
zrax, line 266:
```suggestion
                                ['--rm', '--name', outfile,
```

scripts/pymultic
```diff
@@ -223,14 +255,38 @@ def container_compile(ver, infile):
...
+            shutil.copyfile(comptmp + 'o', outfile)
+            os.unlink(comptmp + 'o')
+        elif os.path.exists(comptmp + 'c'):
+            shutil.copyfile(comptmp + 'c', outfile)
+            os.unlink(comptmp + 'c')
+        os.unlink(comptmp)
+    else:
+        # The easy way
+        proc = subprocess.Popen([container_exe, 'run'] + CONTAINER_EXE_EXTRA_ARGS[container_exe] +
+                                ['--rm', '--name', '{}'.format(outfile),
```
zrax, line 283:
```suggestion
                                ['--rm', '--name', outfile,
```

scripts/pymultic
```diff
@@ -142,6 +156,35 @@ def acquire_python(snekdir, version):
...
+def build_python_container(snekdir, version):
+    realver = PYVERS[version]
+    snek = 'Python-{}'.format(realver)
+    builddir = os.path.join(snekdir, snek)
+    container_exe = get_container_exe()
+    py_container_tag = 'python:{}'.format(realver)
+
+    if subprocess.call([container_exe, 'image', 'inspect', py_container_tag],
+                       stdout=subprocess.DEVNULL,
+                       stderr=subprocess.DEVNULL) == 0:
```
zrax, line 168:
```suggestion
                        stdout=subprocess.DEVNULL,
                        stderr=subprocess.DEVNULL) == 0:
```

scripts/pymultic
```diff
@@ -223,14 +255,38 @@ def container_compile(ver, infile):
...
-                             'python', '-c',
-                             "import py_compile; py_compile.compile('/indir/{}', '{}')".format(infile, outfile)])
-    proc.communicate()
+    if ver in {'1.0', '1.1', '1.2', '1.3', '1.4'}:
+        # The hard way -- hope your code is safe...
+        srcdir = os.path.dirname(os.path.realpath(infile))
+        comptmp = os.path.join(indir, 'pymc_temp.py')
+        if os.path.exists(comptmp):
+            os.unlink(comptmp)
+        shutil.copyfile(infile, comptmp)
```
zrax, line 264: `infile` was modified in this function to be its basename (line 247), meaning this will fail for a relative path.

scripts/pymultic
```diff
@@ -223,14 +255,38 @@ def container_compile(ver, infile):
...
+        os.unlink(comptmp)
+    else:
+        # The easy way
+        proc = subprocess.Popen([container_exe, 'run'] + CONTAINER_EXE_EXTRA_ARGS[container_exe] +
+                                ['--rm', '--name', '{}'.format(outfile),
+                                 '-v', '{}:/indir:Z'.format(indir),
+                                 '-v', '{}:/outdir:Z'.format(os.getcwd()), '-w', '/outdir',
+                                 'python:{}'.format(fullver),
+                                 'python', '-c',
+                                 "import py_compile; py_compile.compile('/indir/{}', '{}')".format(infile, outfile)])
```
zrax, line 288: I assume we have to set the working directory for Python in order for this to work...

scripts/pymultic
dotjrich: Fixed!

scripts/pymultic
dotjrich: Fixed!

scripts/pymultic
dotjrich: Fixed!

scripts/pymultic
dotjrich: Ooh, good catch. I've updated this.

scripts/pymultic
dotjrich: Ah, yeah. Another good catch! Updated.

scripts/pymultic
dotjrich: Yeah, this was an interesting find. I'm guessing that it's just a 2.x -> 3.x transitional oddity. Running `make install` from the 3.0 release installs the binary as `python3` and gives you a warning about not wanting to trample on the 2.x `python` binary. It goes on to say that if you want this installed as `python`, then you need to run `make fullinstall`. That target doesn't seem to exist in any of the other 3.x releases... `make install` does the trick there.

zrax (approved):
Looks great, thanks again!

scripts/pymultic
Hoikas: The `make install` from 3.0 is now `make altinstall`, FWIW.

Issue #224, opened 2021-12-01
Inconsistent error log location and format


CaledoniaProject:
I just noticed pycdc does not have a uniformed logger,

Some of those errors does not start with a comment, e.g

```
default:
            fprintf(stderr, "Unsupported opcode: %s\n", Pyc::OpcodeName(opcode & 0xFF));
            cleanBuild = false;
            return new ASTNodeList(defblock->nodes());
        }
```

This seems to be the only one starts with a comment:

```
if (!cleanBuild || !part1clean) {
        start_line(cur_indent);
        fputs("# WARNING: Decompyle incomplete\n", pyc_output);
    }
```

Why won't you place a `#` before all those logs?

zrax:
There's a key difference between the two:  One is output to stderr (which means it won't get redirected to your output file, and therefore doesn't need to be valid Python syntax) and the other is output to the output stream (i.e. your output file, which must be Python compatible).

CaledoniaProject:
Thanks!

PR #225, opened 2021-12-01
Update ASTree.cpp

CaledoniaProject:
I'm getting `Unsupported opcode: RERAISE` on Python 3.9 constantly, to reduce human works I had to prefix the error logs with a comment.

Let's change this behavior in the decompiler code instead.


zrax:
See my comments in #224 -- the outputs to stderr won't get redirected to your file, and are therefore not required to be valid Python syntax.

CaledoniaProject:
Understood.

PR #245, opened 2022-03-24
Changes in 3.10 under BPO-27129: "Use instruction offset, rather than bytecode offset"

ncaklovic:
3.10 changed all jumps offsets.

See [BPO-27129](https://bugs.python.org/issue27129) and [changes in compile.c](https://github.com/python/cpython/commit/fcb55c0037baab6f98f91ee38ce84b6f874f034a)

This will fix pycdc, not clear what pycdas should show (as operand value).


zrax:
Thanks!  If you want to add a fix for pycdas, I think the only place you'd need to change would be in bytecode.cpp, to correct the "(to <address>)" value in the argument display for jump offset arguments.

zrax (approved):
Looks great, thanks!

PR #247, opened 2022-03-26
Create c-cpp.yml

MyNameJeff12123:


zrax:
There are already CI builds in this repo, for both Linux (GCC) and Windows (MSVC).

PR #249, opened 2022-03-29
Support for conditional expression (if-expression)

ncaklovic:
Support Python [conditional expressions](https://docs.python.org/3/reference/expressions.html#conditional-expressions), code like:

`a = 1 if b else 2`

or

`f(x if b else y)`


ASTree.cpp
```diff
@@ -29,6 +29,43 @@ static bool printDocstringAndGlobals = false;
 /* Use this to keep track of whether we need to print a class or module docstring */
 static bool printClassDocstring = true;
 
+// shortcut for all top/pop calls
+PycRef<ASTNode> StackPopTop(FastStack& stack)
```
zrax, line 33:
```suggestion
static PycRef<ASTNode> StackPopTop(FastStack& stack)
```

ASTree.cpp
```diff
@@ -1363,19 +1402,9 @@ PycRef<ASTNode> BuildFromCode(PycRef<PycCode> code, PycModule* mod)
...
-                if ((curblock->blktype() == ASTBlock::BLK_WHILE
-                            && !curblock->inited())
-                        || (curblock->blktype() == ASTBlock::BLK_IF
-                            && curblock->size() == 0)) {
-                    PycRef<PycObject> fakeint = new PycInt(1);
-                    PycRef<ASTNode> truthy = new ASTObject(fakeint);
-
-                    stack.push(truthy);
-                    break;
-                }
```
zrax, line 1375: Why is this block removed?  It doesn't look related to the change...

ASTree.cpp
```diff
@@ -3256,6 +3285,26 @@ void print_src(PycRef<ASTNode> node, PycModule* mod)
...
+            fputs("( ", pyc_output);
+            print_src(ternary->if_expr(), mod);
+            const auto if_block = ternary->if_block().require_cast<ASTCondBlock>();
+            fputs(" if ", pyc_output);
+            if (if_block->negative())
+                fputs("not ", pyc_output);
+            print_src(if_block->cond(), mod);
+            fputs(" else ", pyc_output);
+            print_src(ternary->else_expr(), mod);
+            fputs(" )", pyc_output);
```
zrax, line 3305:
```suggestion
            fputs(")", pyc_output);
```

zrax (commented):
Please also add at least one test for ternary expressions

ASTree.cpp
ncaklovic: This was "fake" (fakeint, truthy) handling of the same use case. Should be removed.

ncaklovic:
Is there anything else that you think should be changed?

ASTree.cpp
```diff
@@ -3287,13 +3287,16 @@ void print_src(PycRef<ASTNode> node, PycModule* mod)
...
         {
-            /* parenthesis might not be needed,
-             * but when if-expr is part of numerical expression, ternary has the LOWEST precedence
+            /* parenthesis might be needed
+             * 
+             * when if-expr is part of numerical expression, ternary has the LOWEST precedence
              *     print(a + b if False else c)
              * output is c, not a+c (a+b is calculated first)
+             * 
+             * but, let's not add parenthesis - to keep the source as close to original as possible in most cases
```
zrax, line 3296: To clarify, I'm fine with adding parentheses here (ideally we could try to detect if they're necessary, but that's more work).  My suggestion was just to remove the extra whitespace inside the parens, since it's inconsistent with parenthesis output everywhere else.

ASTree.cpp
zrax: Hmm, looking at it, I assumed it was for cases like `if True` or `while 1`, but I can't seem to trigger it...  So I'll assume you're right, since I don't remember the origin of this code :)

zrax (approved):
Looks fine for now.  Thanks for the contribution, and sorry for the delay (I don't have much time to work on pycdc these days, as is probably quite apparent ;) )

PR #265, opened 2022-06-15
Code obj support

clubby789:
For both tools:
- Adds a boolean `-c` flag and a string `-v` paramater for specifying 'this file is a code object' and the Python version respectively
- Adds a new method to load a `PycModule` from a code object (skips header loading and manually sets magic from major/minor)
- Add an `INVALID` variant to the `PycMagic` enum for an invalid magic value

### DIFF
--- a/pyc_module.cpp
+++ b/pyc_module.cpp
@@ -202,6 +202,23 @@ void PycModule::loadFromFile(const char* filename)
     m_code = LoadObject(&in, this).require_cast<PycCode>();
 }
 
+void PycModule::loadFromMarshalledFile(const char* filename, int major, int minor) {
+    PycFile in (filename);
+    if (!in.isOpen()) {
+        fprintf(stderr, "Error opening file %s\n", filename);
+        return;
+    }
+    PycMagic magic = version_to_magic(major, minor);
+    if (magic == PycMagic::INVALID) {
+        fprintf(stderr, "Unsupported version\n");
+        return;
+    }
+    setVersion(magic);
+    m_maj = major;
+    m_min = minor;
+    m_code = LoadObject(&in, this).require_cast<PycCode>();
+}
+
 PycRef<PycString> PycModule::getIntern(int ref) const
 {
     if (ref < 0)
@@ -227,3 +244,51 @@ PycRef<PycObject> PycModule::getRef(int ref) const
         throw std::out_of_range("Ref index out of range");
     return *it;
 }
+
+#define enumVariant(mj, mn) case mn: \
+    return PycMagic::MAGIC_ ## mj ## _ ## mn
+
+PycMagic version_to_magic(int major, int minor) {
+    switch (major) {
+        case 1:
+        switch (minor) {
+            enumVariant(1, 0);
+            case 1:
+            case 2:
+            return PycMagic::MAGIC_1_1;
+            enumVariant(1, 3);
+            enumVariant(1, 4);
+            enumVariant(1, 5);
+            enumVariant(1, 6);
+            default: return PycMagic::INVALID;
+        }
+        case 2:
+            switch (minor) {
+            enumVariant(2, 0);
+            enumVariant(2, 1);
+            enumVariant(2, 3);
+            enumVariant(2, 4);
+            enumVariant(2, 5);
+            enumVariant(2, 6);
+            enumVariant(2, 7);
+            default: return PycMagic::INVALID;
+        }
+        case 3:
+            switch (minor) {
+            enumVariant(3, 0);
+            enumVariant(3, 1);
+            enumVariant(3, 3);
+            enumVariant(3, 4);
+            enumVariant(3, 5);
+            enumVariant(3, 6);
+            enumVariant(3, 7);
+            enumVariant(3, 8);
+            enumVariant(3, 9);
+            enumVariant(3, 10);
+            default: return PycMagic::INVALID;
+        }
+        default:
+            return PycMagic::INVALID;
+        }
+    }
+#undef enumVariant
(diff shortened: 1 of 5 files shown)


pyc_module.cpp
```diff
@@ -202,6 +202,23 @@ void PycModule::loadFromFile(const char* filename)
     m_code = LoadObject(&in, this).require_cast<PycCode>();
 }
 
+void PycModule::loadFromMarshalledFile(const char* filename, int major, int minor) {
```
zrax, line 205:
```suggestion
void PycModule::loadFromMarshalledFile(const char* filename, int major, int minor)
{
```

pyc_module.cpp
```diff
@@ -202,6 +202,23 @@ void PycModule::loadFromFile(const char* filename)
...
+        return;
+    }
+    PycMagic magic = version_to_magic(major, minor);
+    if (magic == PycMagic::INVALID) {
+        fprintf(stderr, "Unsupported version\n");
+        return;
+    }
+    setVersion(magic);
+    m_maj = major;
+    m_min = minor;
```
zrax, line 218: `setVersion()` already sets these...  Additionally, since we're parsing the version directly, we may want to have a way to indicate that Unicode mode is being used for Python 1.6-2.7.  Perhaps checking for a 'u' suffix to the version number would be sufficient?

pyc_module.cpp
```diff
@@ -227,3 +244,51 @@ PycRef<PycObject> PycModule::getRef(int ref) const
...
+            enumVariant(3, 8);
+            enumVariant(3, 9);
+            enumVariant(3, 10);
+            default: return PycMagic::INVALID;
+        }
+        default:
+            return PycMagic::INVALID;
+        }
+    }
+#undef enumVariant
```
zrax, line 294: Newline at EOF

pyc_module.cpp
```diff
@@ -227,3 +244,51 @@ PycRef<PycObject> PycModule::getRef(int ref) const
         throw std::out_of_range("Ref index out of range");
     return *it;
 }
+
+#define enumVariant(mj, mn) case mn: \
+    return PycMagic::MAGIC_ ## mj ## _ ## mn
+
+PycMagic version_to_magic(int major, int minor) {
```
zrax, line 251:
```suggestion
PycMagic version_to_magic(int major, int minor)
{
```

pyc_module.cpp
```diff
@@ -227,3 +244,51 @@ PycRef<PycObject> PycModule::getRef(int ref) const
...
+            enumVariant(3, 7);
+            enumVariant(3, 8);
+            enumVariant(3, 9);
+            enumVariant(3, 10);
+            default: return PycMagic::INVALID;
+        }
+        default:
+            return PycMagic::INVALID;
+        }
+    }
```
zrax, line 293:
```suggestion
    }
}
```

The indentation is a bit inconsistent in this switch, which I think caused the ending braces to be misaligned...
Generally, the pycdc style is to keep `case` aligned with the `switch`:
```c++
switch (<cond>) {
case <N>:
    <stuff>
    break;
case <M>:
    {
        <stuff>
    }
    break;
default:
    <stuff>
    break;
}
```

pycdas.cpp
```diff
@@ -247,11 +250,22 @@ int main(int argc, char* argv[])
...
+                return 1;
+            }
         } else if (strcmp(argv[arg], "--help") == 0 || strcmp(argv[arg], "-h") == 0) {
             fprintf(stderr, "Usage:  %s [options] input.pyc\n\n", argv[0]);
             fputs("Options:\n", stderr);
-            fputs("  -o <filename>   Write output to <filename> (default: stdout)\n", stderr);
-            fputs("  --help          Show this help text and then exit\n", stderr);
+            fputs("  -o <filename>  Write output to <filename> (default: stdout)\n", stderr);
+            fputs("  -c             Specify loading a compiled code object. Requires the version to be set\n", stderr);
+            fputs("  -v <x.y>       Specify a Python version\n", stderr);
```
zrax, line 267:
```suggestion
            fputs("  -v <x.y>       Specify a Python version for loading a compiled code object\n", stderr);
```

pycdc.cpp
```diff
@@ -25,11 +27,22 @@ int main(int argc, char* argv[])
...
+                return 1;
+            }
         } else if (strcmp(argv[arg], "--help") == 0 || strcmp(argv[arg], "-h") == 0) {
             fprintf(stderr, "Usage:  %s [options] input.pyc\n\n", argv[0]);
             fputs("Options:\n", stderr);
-            fputs("  -o <filename>   Write output to <filename> (default: stdout)\n", stderr);
-            fputs("  --help          Show this help text and then exit\n", stderr);
+            fputs("  -o <filename>  Write output to <filename> (default: stdout)\n", stderr);
+            fputs("  -c             Specify loading a compiled code object. Requires the version to be set\n", stderr);
+            fputs("  -v <x.y>       Specify a Python version\n", stderr);
```
zrax, line 44:
```suggestion
            fputs("  -v <x.y>       Specify a Python version for loading a compiled code object\n", stderr);
```

pyc_module.cpp
clubby789: Isn't the `CO_FUTURE_UNICODE_LITERALS` flag set within the code object itself, and wouldn't need to be explicitly added?

zrax: The unicode bytecode flag is added when Python 2.x is run with the (undocumented) `-U` flag, and is not related to the `CO_FUTURE_UNICODE_LITERALS` code flag -- the two can be present or absent independently of each other...

However, from looking at the code, it looks like the only place we actually use `m_unicode` is when printing the Python version in the comments...  There may be something missing there that will need to be addressed eventually, but since this flag was rarely used and doesn't have any meaningful impact on the output, I suppose this can be ignored for now.

PR #266, opened 2022-06-21
feat[junk-code]support remove jump-abs junk

serfend:


zrax:
Hmm, I'm afraid I don't understand what this is fixing...  Is there an example you can provide (ideally as a test case) of this jump "junk"?

serfend:
[pyc.zip](https://github.com/zrax/pycdc/files/8953710/pyc.zip)
we can try this `prob.pyc` which contains abs-jump to invalid address opreation.

PR #270, opened 2022-07-09
3.8: parse for loops and fix ternary ifs

ahaensler:
The PR takes a small step and adds support for loops >3.8.
Please find a test case below.
```
if 1 == 1:
    a = [x for x in "123"]
if 2 == 2:
    for a in "456":
        print("loop")
```

Output before PR
```
return None
```

Output after PR
```
if 1 == 1:
    a = (lambda .0: [ x for x in .0 ])('123')
if 2 == 2:
    for a in '456':
        print('loop')
return None
```


ASTNode.h
```diff
@@ -550,13 +550,14 @@ class ASTCondBlock : public ASTBlock {
 
 class ASTIterBlock : public ASTBlock {
 public:
-    ASTIterBlock(ASTBlock::BlkType blktype, int end, PycRef<ASTNode> iter)
-        : ASTBlock(blktype, end), m_iter(std::move(iter)), m_idx(), m_comp() { }
+    ASTIterBlock(ASTBlock::BlkType blktype, int start, int end, PycRef<ASTNode> iter)
```
ahaensler, line 553: Save the position of the for loop and use it to find block extents. Later, continue statements could be inferred

ASTree.cpp
```diff
@@ -54,7 +54,11 @@ static void CheckIfExpr(FastStack& stack, PycRef<ASTBlock> curblock)
     if (curblock->nodes().size() < 2)
         return;
     auto rit = curblock->nodes().crbegin();
-    ++rit; // the last is "else" block, the one before should be "if" (could be "for", ...)
+    // the last is "else" block, the one before should be "if" (could be "for", ...)
+    if ((*rit)->type() != ASTNode::NODE_BLOCK ||
```
ahaensler, line 58: Checks exactly what the comment says now. This had false positives for ternary ifs

zrax (commented):
Thanks for the contribution!  The only thing I would ask is to please add a test case for it (at least for the loops that succeed)

ahaensler:
I have added a test. With new compilers, there's always `return None` at the end. Maybe I'll look into it later

PR #279, opened 2022-09-14
Added recursion depth limiting

blincoln-bf:
Decompyle++ is an amazing tool, but it will sometimes get stuck in infinite recursion loops until the system runs out of memory.

I've added recursion-limiting on both object loading and source code generation to work around this issue when analyzing problematic code. It defaults to 100 levels (probably more than enough for any real-world code), but can be changed with the new -m command-line option.

This doesn't fix the underlying issue that causes a particular infinite recursion, but it usually allows partial decompilation of the problem code. More importantly, if Decompyle++ is being used in a batch script, explicitly failing a single run of the tool and then proceeding to the next file is preferable to the entire script failing in the middle when one problem file causes infinite recursion.

### DIFF
--- a/ASTree.cpp
+++ b/ASTree.cpp
@@ -2604,27 +2604,27 @@ static int cmp_prec(PycRef<ASTNode> parent, PycRef<ASTNode> child)
 }
 
 static void print_ordered(PycRef<ASTNode> parent, PycRef<ASTNode> child,
-                          PycModule* mod)
+                          PycModule* mod, int currentDepth, int maxDepth)
 {
     if (child.type() == ASTNode::NODE_BINARY ||
         child.type() == ASTNode::NODE_COMPARE) {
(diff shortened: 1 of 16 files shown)


blincoln-bf:
@zrax, any interest in merging this? I've been working on some automation that needs a Decompyle++ that fails gracefully instead of running out of memory, but I can keep it in the existing fork instead.

zrax:
The concept of the change seems like a good one, but this is hard to review in its current state...  There are a bunch of extra "noise" changes, including indentation and style changes which are unrelated to the actual work.  Also, I wonder if there's a better way to track the recursion depth than adding two extra parameters to almost every function in the code base :)

blincoln-bf:
OK. I can direct people to my fork for now, and if you add an equivalent feature later, I'll update the documentation to point back to the main repo.

PR #303, opened 2022-12-11
Fix for Unsupported opcode: WITH_EXCEPT_START

cw2k:
-> Fixed for
Unsupported opcode: [WITH_EXCEPT_START](https://docs.python.org/3/library/dis.html#opcode-WITH_EXCEPT_START)
and
Unsupported opcode: [RERAISE](https://docs.python.org/3/library/dis.html#opcode-RERAISE)
 by just adding these into the AST builder but without implementing any handler for it.
( too complicated for me to understand the code and types )
However now it runs.

-> added some useful comments on some opcodes
-> make the  BLOCK_DEBUG to also show Opcodes

### DIFF
--- a/ASTree.cpp
+++ b/ASTree.cpp
@@ -1765,6 +1765,7 @@ PycRef<ASTNode> BuildFromCode(PycRef<PycCode> code, PycModule* mod)
                 }
             }
             break;
+
         case Pyc::POP_EXCEPT:
             /* Do nothing. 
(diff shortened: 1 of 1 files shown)


phdphuc:
Thanks this fixes [RERAISE](https://docs.python.org/3/library/dis.html#opcode-RERAISE) and [WITH_EXCEPT_START](https://docs.python.org/3/library/dis.html#opcode-WITH_EXCEPT_START) for me

PR #312, opened 2023-01-19
Add GET_YIELD_FROM_ITER support

TheHellTower:


zrax:
Hi, thanks for the contribution!

Unfortunately, it looks like your editor has massively refactored/re-styled ASTree.cpp, which means I cannot review it in its current state.  Please restore the original style of the file, and then only update the parts of the code which you have modified, to make reviews and merging easier.

TheHellTower:
> Hi, thanks for the contribution!
> 
> Unfortunately, it looks like your editor has massively refactored/re-styled ASTree.cpp, which means I cannot review it in its current state. Please restore the original style of the file, and then only update the parts of the code which you have modified, to make reviews and merging easier.

Hi, yeah I think the indentation had some problems in the original project or the editor messed with it.
I will do it right now.

TheHellTower:
Updated

zrax:
It looks like this also fixes the previously-expected failures in the `yield_from` test for Python 3.5 and 3.7.  You can mark these as passing by moving them from `tests/xfail/` to `tests/compiled/`.  In fact, it doesn't look like the new test (`GET_YIELD_FROM_ITER.py`) tests anything that isn't already in the existing `yield_from` test (and others), so perhaps the new test case isn't necessary?  You can add a Python 3.9 version of the .pyc if you want to ensure it's covered too.

TheHellTower:
> compiled

Well I didn't see there was a test for the `yield_from` I directly tried myself, so yeah for this missed detail the new test case isn't really necessary but I don't think it does matter too much it was just for testing purpose so I will delete the files I added and compile the existing test for 3.9, I will move the files right now too.

TheHellTower:
Everything should be fine now !

TheHellTower:
I had fixed GEN_START too so I will check if there is a test for it and open a PR as soon as this completely finished and merged.

TheHellTower:
Apparently no test exist and I'm not really good with python so I don't know how to make a code using it, if you have one don't hesitate to send it to me so I can make a PR with the tests, compiled & tokenized example.

PR #324, opened 2023-02-26
Add `PUSH_NULL`, `PRECALL_A`, `CALL_A` support

### LINKED ISSUE #314: Unsupported opcode: PRECALL
# WARNING: Decompyle incomplete
Unsupported opcode: PRECALL

### LINKED ISSUE #322: PUSH_NULL
====> [Hi.zip] (https://github.com/zrax/pycdc/files/10797905/Hi.zip)

# File: hi.pyc (Python 3.11)

Unsupported opcode: PUSH_NULL
import marshal
# WARNING: Decompyle incomplete

How to fix this Help

TheHellTower:
Closes #314 
Closes #322

**`Before merging this PR I suggest a review`**

I used #322 sample

<br>


ASTree.cpp
```diff
@@ -2556,6 +2556,94 @@ PycRef<ASTNode> BuildFromCode(PycRef<PycCode> code, PycModule* mod)
         case Pyc::RESUME_A:
             /* Treated as no-op for decompyle purposes */
             break;
+        case Pyc::PUSH_NULL:
+            {
+                stack.push(NULL);
+            }
```
zrax, line 2562:
```suggestion
            stack.push(nullptr);
```

ASTree.cpp
```diff
@@ -2556,6 +2556,94 @@ PycRef<ASTNode> BuildFromCode(PycRef<PycCode> code, PycModule* mod)
...
+            break;
+        case Pyc::PRECALL_A:
+            /* Treated as no-op for decompyle purposes */
+            break;
+        case Pyc::CALL_A:
```
zrax, line 2567: If this is identical to (or almost identical to) the `Pyc::CALL_FUNCTION_A` case, they should be combined rather than copied.

zrax (commented):
In addition, please add at least one unit test to validate the new opcodes

TheHellTower:
> In addition, please add at least one unit test to validate the new opcodes

I will add one when everything will be solved for `Pyc::CALL_A` that is almost identical to `Pyc::CALL_FUNCTION_A` that is

ASTree.cpp
zrax: You can add a conditional on the opcode as has been done in a few other places within shared case branches...  For example, something like:

```c++
        ...
        stack.pop();
        if (opcode == Pyc::CALL_A)
            stack.pop();
        ...
```

TheHellTower: Yeah I thought about it but not exactly about a conditional yeah I will modify this right now on my side and push it

zrax:
Yes, looks much cleaner now, thank you.

zrax:
That sample isn't really a great unit test, unfortunately...  I looked to see if we already had a unit test that just checks basic function calls, but I didn't see one.  I feel like that's already missing from the current test cases, so I added #325, which I verified does generate all of `CALL`, `PRECALL` and `PUSH_NULL` on Python 3.11.

TheHellTower:
> Reference in new is

Well, should I delete the tests I added then ?

TheHellTower:
I believe this CodeQL fail is not related to me, apparently it's using a outdated version of CodeQL Actions.

TheHellTower:
the `.` at the end make it reach a 404 page so

TheHellTower:
It is ready to be merged as soon as you are ready to do it

zrax (commented):
I also tested this with the newly added unit test (test_calls.py), but it is currently failing due to the second `sys.stdout.write` being replaced by `None`.  Please include a compiled Python 3.11 version of the test file, and see if you can either fix the issue, or at least identify what causes it (if it's unrelated to this change)

ASTree.cpp
```diff
@@ -2548,13 +2545,17 @@ PycRef<ASTNode> BuildFromCode(PycRef<PycCode> code, PycModule* mod)
             variable_annotations = true;
             break;
         case Pyc::CACHE:
+        case Pyc::GET_ITER:
+        case Pyc::GET_YIELD_FROM_ITER:
+        case Pyc::PRECALL_A:
+        case Pyc::RESUME_A:
```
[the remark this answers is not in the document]
TheHellTower: So I should Put the 4 other cases in their own place with their own comment ?

TheHellTower:
> I also tested this with the newly added unit test (test_calls.py), but it is currently failing due to the second `sys.stdout.write` being replaced by `None`. Please include a compiled Python 3.11 version of the test file, and see if you can either fix the issue, or at least identify what causes it (if it's unrelated to this change)

I will try that right now

TheHellTower:
If I remove this conditional https://github.com/zrax/pycdc/pull/324#discussion_r1119174515 then the result make the second print being replaced by `None` I'm not really sure what could be the issue I will try to check how this can be solved but I might not be able to solve it

zrax:
Looking at the Python 3.11 source code (Python/ceval.c), it looks like the extra NULL on the stack (from `PUSH_NULL`) is used to indicate whether the `CALL` is a method or function call.  More specifically, it conditionally pops the null only when it's present, so we'll need to do the same thing here.

TheHellTower:
> Looking at the Python 3.11 source code (Python/ceval.c), it looks like the extra NULL on the stack (from `PUSH_NULL`) is used to indicate whether the `CALL` is a method or function call. More specifically, it conditionally pops the null only when it's present, so we'll need to do the same thing here.

So I need to verify if a `PUSH_NULL` is present `3 indexes` before the `CALL` and if yes I pop at this moment ?

TheHellTower:
It does work on my side I added one more check: `stack.top() == nullptr`

Issue #294, opened 2022-11-04
Unicode support?


i486:
Other decompilers output
print('共比對0筆Log')

pycdc's output
print('\xe5\x85\xb1\xe6\xaf\x94\xe5\xb0\x8d0\xe7\xad\x86Log')

Are there any command line switch or something to output actual strings?

Not sure if it's an issue but i am using Windows build.

Issue #121, opened 2017-04-16
bug in unicode


TechComet:
Welcome

if string in unicode (Arabic) show string "\xd8\xb1\xd8\xb3\xd8\xa7\xd9\x84\xd8\xa9"

please fix this

rocky:
Would you attach the smallest Python program that exhibits this problem. Also, what version of Python is the bytecode in. Thanks.

PR #344, opened 2023-04-27
Fix unicode string display bug

seashell11234455:


zrax:
Fixes #121 
Fixes #294

PR #348, opened 2023-05-29
Add `BINARY_OP` support

### LINKED ISSUE #333: Unsupported opcode: BINARY_OP

TheHellTower:
Closes #333

[](https://i.imgur.com/VP2GUbT.mp4)


TheHellTower:
I literally compiled from Python 3.11 and used Python 3.11 to launch the token_dump(even if it might be useless) so I don't get what is going wrong there:

brave_KfBtjcvdxr

ASTree.cpp
```diff
@@ -157,6 +157,7 @@ PycRef<ASTNode> BuildFromCode(PycRef<PycCode> code, PycModule* mod)
         }
 
         switch (opcode) {
+        case Pyc::BINARY_OP_A:
```
zrax, line 160: The `_A` implies that there is an operand; in this case, the operand is the binary operation to perform (it's not just for addition).  You'll need to add a whole separate case for this.

tests/input/BINARY_OP.py
```diff
@@ -0,0 +1,4 @@
+def add(a, b):
+    return a + b
+
+add(1, 3)
```
zrax, line 4: Probably doesn't need to include a function to generate the opcode(s) -- might as well get rid of the function to reduce what's being tested.  Also, please ensure files end with a newline.

tests/input/BINARY_OP.py
```diff
@@ -0,0 +1,4 @@
+def add(a, b):
+    return a + b
```
zrax, line 2: Please test all cases of binary ops, not just addition (that would have caught the issue above)

tests/tokenized/BINARY_OP.txt
```diff
@@ -0,0 +1,5 @@
+def add ( a , b ) : <EOL>
+<INDENT>
+return a + b <EOL>
+<OUTDENT>
+add ( 1 , 3 ) <EOL>
```
zrax, line 5: This file is failing the diff on Linux because it has Windows (CR+LF) line endings, which the Linux `diff` tool doesn't understand.  You can use something like dos2unix on  it, or configure your editor to save it with Linux (LF only) line endings.

zrax (changes requested):

TheHellTower:
Thanks for all the comments, I will try to figure this out, but for the OPCode [BINARY_OP#3.11](https://docs.python.org/3.11/library/dis.html#opcode-BINARY_OP) I don't get the "depending on the value of op" and can't find what it's supposed to mean and use.
Is the argument supposed to be the whole operation ?

Will the BINARY_OP be big ? So I know if I will need some time for it or if it will be "fast"

TheHellTower:
brave_TAiFaRLnh8
Well I see what you mean now !
I will check this deeply :)

zrax:
You can find the list of binary operations used in the opcode in Python's source:
https://github.com/python/cpython/blob/3.11/Include/opcode.h#L200

Although these don't map directly to the values in ASTBinary, you could probably generate a simple lookup, and then create the appropriate ASTBinary node with the mapped operation.

TheHellTower:
> You can find the list of binary operations used in the opcode in Python's source: https://github.com/python/cpython/blob/3.11/Include/opcode.h#L200
> 
> Although these don't map directly to the values in ASTBinary, you could probably generate a simple lookup, and then create the appropriate ASTBinary node with the mapped operation.

Ah thanks ! It's helpful ! I didn't think about it ! :)

TheHellTower:
brave_7rgLEW03OV
How ? I used dos2unix(With GIT Bash)

explorer_fL90eS0auF
brave_cduOtmPEhX
So close...

zrax:
You don't need to run dos2unix on the source files, just the one .txt file that was causing diff issues...  Regarding the remaining differences, it's likely that the Python compiler is optimizing things for you.  Note that the test files don't actually need to run, just compile -- so you can change the code to prevent the optimizations (such as using variables instead of constants, etc)

ASTNode.h
```diff
@@ -1,735 +1,794 @@
...
+        case 22:
+            return BIN_IP_RSHIFT;
+        case 23:
+            return BIN_IP_SUBTRACT;
+        case 24:
+            return BIN_IP_DIVIDE;
+        case 25:
+            return BIN_IP_XOR;
+        default:
+            return BIN_ADD; // Default value if operand is out of range
```
zrax, line 196: I'd prefer to add an error case here, rather than falling back to addition.  As a suggestion, you could add a `BIN_INVALID` value to the enum, and then print an error in the `BINARY_OP_A` case if this returned `BIN_INVALID`

ASTNode.h
```diff
@@ -1,735 +1,794 @@
...
+        BIN_ATTR, BIN_POWER, BIN_MULTIPLY, BIN_DIVIDE, BIN_FLOOR, BIN_MODULO,
+        BIN_ADD, BIN_SUBTRACT, BIN_LSHIFT, BIN_RSHIFT, BIN_AND, BIN_XOR,
+        BIN_OR, BIN_LOG_AND, BIN_LOG_OR, BIN_MAT_MULTIPLY,
+        /* Inplace operations */
+        BIN_IP_ADD, BIN_IP_SUBTRACT, BIN_IP_MULTIPLY, BIN_IP_DIVIDE,
+        BIN_IP_MODULO, BIN_IP_POWER, BIN_IP_LSHIFT, BIN_IP_RSHIFT, BIN_IP_AND,
+        BIN_IP_XOR, BIN_IP_OR, BIN_IP_FLOOR, BIN_IP_MAT_MULTIPLY,
+    };
+
+    static BinOp getBinOpFromOperand(int operand) {
```
zrax, line 141: This function is long enough it's probably worth moving the implementation to ASTNode.cpp rather than including it inline.

TheHellTower:
> You don't need to run dos2unix on the source files, just the one .txt file that was causing diff issues... Regarding the remaining differences, it's likely that the Python compiler is optimizing things for you. Note that the test files don't actually need to run, just compile -- so you can change the code to prevent the optimizations (such as using variables instead of constants, etc)

Well yeah I ran it on the txt file but also the fact it change the `[]` to `()` doesn't impact it (even thought it's not the same opcode that do that I tried without modifications) ? And which code would you recommend then because That's the simplest one I got..

ASTNode.h
TheHellTower: True it was mostly for testing I will update this

ASTNode.h
TheHellTower: Resolved locally !

TheHellTower:
I moved the function definition + added `BIN_INVALID`

ASTNode.cpp
```diff
@@ -1,81 +1,139 @@
...
+
+
+/* ASTBinary */
+const char* ASTBinary::op_str() const
+{
+    static const char* s_op_strings[] = {
+        ".", " ** ", " * ", " / ", " // ", " % ", " + ", " - ",
+        " << ", " >> ", " & ", " ^ ", " | ", " and ", " or ", " @ ",
+        " += ", " -= ", " *= ", " /= ", " %= ", " **= ", " <<= ",
+        " >>= ", " &= ", " ^= ", " |= ", " //= ", " @= ", "<BAD>"
```
zrax, line 34:
```suggestion
        " >>= ", " &= ", " ^= ", " |= ", " //= ", " @= ", " <INVALID> "

```

ASTNode.cpp
```diff
@@ -1,81 +1,139 @@
...
+    };
+    return s_op_strings[op()];
+}
+
+const ASTBinary::BinOp ASTBinary::getBinOpFromOperand(int operand) {
```
zrax, line 39:
```suggestion
ASTBinary::BinOp ASTBinary::getBinOpFromOperand(int operand)
{
```

ASTNode.h
```diff
@@ -1,735 +1,738 @@
...
+    ASTBinary(PycRef<ASTNode> left, PycRef<ASTNode> right, int op,
+              int type = NODE_BINARY)
+        : ASTNode(type), m_op(op), m_left(std::move(left)), m_right(std::move(right)) { }
+
+    PycRef<ASTNode> left() const { return m_left; }
+    PycRef<ASTNode> right() const { return m_right; }
+    int op() const { return m_op; }
+    bool is_inplace() const { return m_op >= BIN_IP_ADD; }
+    virtual const char* op_str() const;
+    static const BinOp getBinOpFromOperand(int operand);
```
zrax, line 152:
```suggestion
    static BinOp getBinOpFromOperand(int operand);
```

ASTree.cpp
```diff
@@ -1,3545 +1,3560 @@
...
+            if (op == ASTBinary::BIN_INVALID) {
+                fprintf(stderr, "Unsupported `BINARY_OP` operand value: %s\n", operand);
+            }
+            else {
+                PycRef<ASTNode> right = stack.top();
+                stack.pop();
+                PycRef<ASTNode> left = stack.top();
+                stack.pop();
+                stack.push(new ASTBinary(left, right, op));
+            }
```
zrax, line 172: In order to avoid breaking the stack for the rest of the decompylation, I'd recommend still putting the ASTBinary on the stack even in the failure case...
```suggestion
            ASTBinary::BinOp op = ASTBinary::getBinOpFromOperand(operand);
            if (op == ASTBinary::BIN_INVALID)
                fprintf(stderr, "Unsupported `BINARY_OP` operand value: %s\n", operand);
            PycRef<ASTNode> right = stack.top();
            stack.pop();
            PycRef<ASTNode> left = stack.top();
            stack.pop();
            stack.push(new ASTBinary(left, right, op));
```

This ends up working fine since we still emit a valid `op_str()` even for `BIN_INVALID`

tests/input/BINARY_OP.py
```diff
@@ -0,0 +1,25 @@
...
+print('Floor Division:', a // b)
+print('Exponentiation:', a ** b)
+print('Greater than:', a > b)
+print('Less than:', a < b)
+print('Greater than or equal to:', a >= b)
+print('Less than or equal to:', a <= b)
+print('Equal to:', a == b)
+print('Not equal to:', a != b)
+print('Logical AND:', True and False)
+print('Logical OR:', True or False)
```
zrax, line 16: To avoid constant folding, try
```suggestion
print('Logical AND:', a and b)
print('Logical OR:', a or b)
```

tests/input/BINARY_OP.py
```diff
@@ -0,0 +1,25 @@
...
+print('Bitwise AND:', a & b)
+print('Bitwise OR:', a | b)
+print('Bitwise XOR:', a ^ b)
+print('Bitwise left shift:', a << b)
+print('Bitwise right shift:', a >> b)
+print('Membership in:', b in [1, 2, 3])
+print('Membership not in:', b not in [1, 2, 3])
```
zrax, line 23: To avoid an optimization, you could change these to tuples, or just use `a in b` and `a not in b`...

ASTree.cpp
TheHellTower: True sorry I didn't think about it that's a good idea actually

TheHellTower:
Okay going to update the sample now

TheHellTower:
Oh wait it broken the code

TheHellTower:
cmd_XoW4LEZPUm

TheHellTower:
The script update produced this but when I remove the handling of `BINARY_OP` (obviously) it's "normal" for pycdc to do this already so one of the OPCode is capricious :eyes:

zrax:
It looks like Python is emitting jumps (another optimization) instead of `BINARY_OP` for the `a and b` and `a or b` cases (and additionally, pycdc is generating the wrong output, but that's a different issue)...  I'm not sure how to convince it to generate `BINARY_OP` here, so I'd say it's fine to comment out those two lines in the test for now.

zrax (approved):
The opcode is fully supported, we just don't have a good way to test it right now.  If you look at the disassembly, you'll see that `BINARY_OP` isn't even being generated for those cases.  However, if it were, we should handle it correctly.

PR #349, opened 2023-06-01
Python bindings & package

percevalw:
Hi,

Thanks for this tool, I've been finding myself needing to recover functions from Python `__code__` objects. To facilitate / speed up this step, I propose this PR, which adds Python bindings to allow using `pycdc` on marshaled objects from Python directly.

The key changes are the following:

- Addition of Python bindings using `pybind11`.
- A new Python package `pycdc`, enabling users to `pip install .`.
- Additional dedicated CMake file, to keep changes to the codebase minimal.
- Conversion of C style IO to C++ streams, to facilitate output redirection to Python.
- Necessary Python build files like `setup.py`, `pyproject.toml`

I've tried to keep changes to the codebase and requirements to a minimum. Please let me know if there are any improvements needed. 

I've not added Github actions at this point but having the package on PyPi would be great. I can add them if the above changes work for you.


zrax:
This sounds like a neat feature...  In order to make it easier to review, I merged the first part (using C++ iostreams) separately, since that has other benefits as well and doesn't appear to impact performance much.  Can you rebase this on the latest master when you get a chance?

percevalw:
Thank you for you efforts ! I've rebased and added an option for decompyling non-marshaled outputs (`decompyle(..., version=None)`, similar to no `-c -v ...`)

.gitignore
```diff
@@ -1,7 +1,43 @@
...
+CMakeFiles/
+Testing/
+CMakeCache.txt
+build.ninja
+.ninja_deps
+.ninja_log
+
+# Executables
+pycdc
+pycdas
```
zrax, line 43: Rather than generating build output in the source tree, why not configure CMake to use a build directory inside Python's build temp dir?

pyc_module.cpp
```diff
@@ -239,6 +239,44 @@ void PycModule::loadFromMarshalledFile(const char* filename, int major, int mino
...
+{
+    if (!isSupportedVersion(major, minor)) {
+        fprintf(stderr, "Unsupported version %d.%d\n", major, minor);
+        return;
+    }
+    m_maj = major;
+    m_min = minor;
+    m_unicode = (major >= 3);
+    m_code = LoadObject(&stream, this).cast<PycCode>();
+}
```
zrax, line 278: `loadFromFile` and `loadFromMarshalledFile` can call these versions with the `PycFile` they open, instead of having to maintain duplicate implementations

percevalw: Good point !

.gitignore
percevalw: Ah, you're right, it's actually just related to my setup since I used the root dir as build output. I'll remove this.

PR #365, opened 2023-07-18
Added support for opcodes: (in DESC)

TiZCrocodile:
`KW_NAMES`
`POP_JUMP_FORWARD_IF_FALSE`
`POP_JUMP_FORWARD_IF_TRUE`
`LOAD_GLOBAL`

now I added the test files.
EDIT: the `LOAD_GLOBAL` opcode exists in the tests_kwnames test files. so no need to check that separately.


ASTree.cpp
```diff
@@ -1029,15 +1067,18 @@ PycRef<ASTNode> BuildFromCode(PycRef<PycCode> code, PycModule* mod)
...
                 int offs = operand;
                 if (mod->verCompare(3, 10) >= 0)
                     offs *= sizeof(uint16_t); // // BPO-27129
                 if (opcode == Pyc::JUMP_IF_FALSE_A
-                        || opcode == Pyc::JUMP_IF_TRUE_A) {
+                        || opcode == Pyc::JUMP_IF_TRUE_A
+                        || opcode == Pyc::POP_JUMP_FORWARD_IF_TRUE_A
+                        || opcode == Pyc::POP_JUMP_FORWARD_IF_FALSE_A) {
                     /* Offset is relative in these cases */
-                    offs = pos + operand;
```
TiZCrocodile, line 1040: here its better to add `pos`, the 2 opcodes `JUMP_IF_FALSE` and `JUMP_IF_TRUE` were anyway only in python 3.0, but since only from python 3.10 the size of the operand is doubled, then you would have to do this line:
`offs = pos + operand * sizeof(uint16_t);`
so just add pos when its relative.

dberrocal-git:
Could you add codes like JUMP_BACKWARD?

Hazed2004:
hey, great stuff, thanks for you contribution

PR #391, opened 2023-10-18
Add support for 3.6+ default arguments

### LINKED ISSUE #138: 3.6 default parameter handling
```
def foo1(bar, baz=1):
    return 1
def foo2(bar, baz, qux=1):
    return 2
def foo3(bar, baz=1, qux=2):
    return 3
def foo4(bar, baz, qux=1, quux=2):
    return 4
```
deocompiles on 3.6 to:

```
def foo1(bar, baz = (1,)):
    return 1


def foo2(bar, baz, qux = (1,)):
    return 2


def foo3(bar, baz, qux = (1, 2)):
    return 3


def foo4(bar, baz, qux, quux = (1, 2)):
    return 4
```

ptrstr:
Fixes #138 and #155 

Python 3.6 moved from pushing the default arguments individually to the stack to pushing a tuple with them.

Modifications:
- [`LOAD_CLOSURE`](https://docs.python.org/3.6/library/dis.html#opcode-LOAD_CLOSURE) should push junk to the stack instead of skipping, as `MAKE_FUNCTION` with operand 8 will have a structure like follows:
```
 10           0 LOAD_CONST               4 (('',))
              2 LOAD_CLOSURE             0 (digestmod)
              4 BUILD_TUPLE              1
              6 LOAD_CONST               2 (<code object <lambda> at 0x5629f4e717a0, file "example.py", line 10>)
              8 LOAD_CONST               3 ('__init__.<locals>.<lambda>')
             10 MAKE_FUNCTION            9
```
- [`MAKE_FUNCTION`](https://docs.python.org/3.6/library/dis.html#opcode-MAKE_FUNCTION) will handle default arguments of type 1, 2 and 8 (4 is about type hints).

### DIFF
--- a/ASTree.cpp
+++ b/ASTree.cpp
@@ -1545,15 +1545,42 @@ PycRef<ASTNode> BuildFromCode(PycRef<PycCode> code, PycModule* mod)
                 }
 
                 ASTFunction::defarg_t defArgs, kwDefArgs;
-                const int defCount = operand & 0xFF;
-                const int kwDefCount = (operand >> 8) & 0xFF;
-                for (int i = 0; i < defCount; ++i) {
-                    defArgs.push_front(stack.top());
-                    stack.pop();
-                }
-                for (int i = 0; i < kwDefCount; ++i) {
-                    kwDefArgs.push_front(stack.top());
-                    stack.pop();
+
+                if (mod->minorVer() <= 5) {
+                    const int defCount = operand & 0xFF;
+                    const int kwDefCount = (operand >> 8) & 0xFF;
+                    for (int i = 0; i < defCount; ++i) {
+                        defArgs.push_front(stack.top());
+                        stack.pop();
+                    }
+                    for (int i = 0; i < kwDefCount; ++i) {
+                        kwDefArgs.push_front(stack.top());
+                        stack.pop();
+                    }
+                } else {
+                    if (operand & 0x08) {
+                        stack.pop();
+                    }
+
+                    if (operand & 0x04) {
+                        stack.pop();
+                    }
+
+                    if (operand & 0x02) {
+                        PycRef<ASTConstMap> defaultsDict = stack.top().cast<ASTConstMap>();
+                        stack.pop();
+
+                        for (PycRef<ASTNode> value : defaultsDict->values())
+                            kwDefArgs.push_front(value);
+                    }
+
+                    if (operand & 0x01) {
+                        PycRef<PycTuple> defaultsTuple = stack.top().cast<ASTObject>()->object().cast<PycTuple>();
+                        stack.pop();
+
+                        for (PycRef<PycObject> value : defaultsTuple->values())
+                            defArgs.push_back(new ASTObject(value));
+                    }
                 }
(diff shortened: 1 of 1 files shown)


ASTree.cpp
```diff
@@ -1469,6 +1469,7 @@ PycRef<ASTNode> BuildFromCode(PycRef<PycCode> code, PycModule* mod)
             break;
         case Pyc::LOAD_CLOSURE_A:
             /* Ignore this */
+            stack.push(new ASTNode());
```
zrax, line 1472: Shouldn't this be pushing a free/cell var, rather than an empty node?

ASTree.cpp
```diff
@@ -1545,15 +1546,42 @@ PycRef<ASTNode> BuildFromCode(PycRef<PycCode> code, PycModule* mod)
...
-                const int kwDefCount = (operand >> 8) & 0xFF;
-                for (int i = 0; i < defCount; ++i) {
-                    defArgs.push_front(stack.top());
-                    stack.pop();
-                }
-                for (int i = 0; i < kwDefCount; ++i) {
-                    kwDefArgs.push_front(stack.top());
-                    stack.pop();
+
+                if (mod->minorVer() <= 5) {
```
zrax, line 1550: This needs to check majorVer as well...  Probably better to use `verCompare`:
```suggestion
                if (mod->verCompare(3, 6) < 0) {
```

zrax (commented):
It would also be good to add or update tests for this case

ptrstr:
> It would also be good to add or update tests for this case

The tests looked a bit strange from my observations (Python code wasn't the same as the expected output).

I rebuilt the `.pyc` files for the versions that were already there and removed the xfail as now it passed.

zrax (commented):
The names of the compiled test files are not correct (unfortunately, it's a bit misleading since the input file is "...`_py3`", so "test_functions_py3.3.0.pyc" means "test_functions_py3.py for Python 3.0").  Also, when I rebuilt the test files for Python 3.4 (which was one of the original cases included), it appears to have a new failure:
```
tests/test_functions_py3.3.4.pyc.tok.txt does not match tests/tokenized/test_functions_py3.txt:
--- tests/tokenized/test_functions_py3.txt      2023-10-20 14:05:18.112904714 -0700
+++ tests/test_functions_py3.3.4.pyc.tok.txt    2023-10-20 14:07:39.786332445 -0700
@@ -58,7 +58,7 @@
 <INDENT>
 pass <EOL>
 <OUTDENT>
-def x6d ( foo = 1 , * , bar = 2 ) : <EOL>
+def x6d ( foo = 2 , * , bar = 'bar' ) : <EOL>
 <INDENT>
 pass <EOL>
 <OUTDENT>
@@ -74,6 +74,6 @@
 <INDENT>
 pass <EOL>
 <OUTDENT>
-def x7d ( foo = 1 , * , bar = 2 , ** kwargs ) : <EOL>
+def x7d ( foo = 2 , * , bar = 'bar' , ** kwargs ) : <EOL>
 <INDENT>
 pass <EOL>
```

Issue #408, opened 2023-12-01
Unsupported Node type: 12


gagmeng:
`E:\PikPak_By_Shimily\Python-Decompiler-main\Python-Decompiler-main>E:\PikPak_By_Shimily\Python-Decompiler-main\Python-Decompiler-main\pycdc.exe "E:\PikPak_By_Shimily\pyinstxtractor-ng-main\pyinstxtractor-ng-main\PikPak Desktop_unpacked.exe_extracted\PYZ-00.pyz_extracted\app\view\main_window.pyc"
# Source Generated with Decompyle++
# File: main_window.pyc (Python 3.9)

'''
Author: qianye
Date: 2023-07-13 17:18:55
LastEditTime: 2023-11-15 08:17:46
Description:
Software: Visual Studio Code
'''
import hashlib
import json
import os
import platform
import sys
import traceback
import urllib.request as urllib
from datetime import datetime
from urllib.parse import quote
import magneturi
import rsa
from PyQt5.QtCore import QMutex, Qt, QTimer, QUrl
from PyQt5.QtGui import QCursor, QDesktopServices, QIcon
from PyQt5.QtWidgets import QApplication, QDialog, QSystemTrayIcon
from qfluentwidgets import Action, DotInfoBadge
from qfluentwidgets import FluentIcon as FIF
from qfluentwidgets import FluentWindow, InfoBadgePosition, InfoBar, InfoBarIcon, InfoBarPosition, MenuAnimationType, MessageBox, NavigationItemPosition, PopupTeachingTip, PushButton, RoundMenu, SystemTrayMenu, TeachingTipTailPosition, TeachingTipView
import resource_rc
from common import encrypt
from common.config import APP_NAME, BUY_VIP_URL, CHECK_UPDATE_URL, RELEASE_URL, VERSION, cfg, mysettings
from common.custom_components import CustomDialog, LoadingBox, MyInfoBar, MyMessageBox, NavigationAvatarWidget, MyFluentIcon
from common.signal_bus import signalBus
from common.singleton import Singleton
from common.subthread import *
from common.token_manager import token_manger
from about_interface import AboutInterface
from activate_dialog import ActivateDialog
from activation_dialog import ActivationDialog
from aria2_dialog import Aria2Dialog
from bitcomet_dialog import BitcometDialog
from connect_dialog import ConnectDialog
from database_dialog import DataBaseDialog
from exception_dialog import ExceptionDialog
from file_interface import FileInterface
[...]

Issue #395, opened 2023-11-01
Unsupported Node type: 27 


zn123:
```
E:\temp_vc\pycdc\build_v141\Release>pycdc.exe model.pyc >model.py
Unsupported Node type: 27
Unsupported Node type: 27
Unsupported Node type: 27
Unsupported argument found for LIST_EXTEND
Unsupported opcode: LIST_TO_TUPLE
Unsupported opcode: CALL_FUNCTION_EX
Unsupported opcode: DICT_MERGE
Unsupported Node type: 27
Unsupported Node type: 27
Unsupported argument found for LIST_EXTEND
Unsupported opcode: LIST_TO_TUPLE
Unsupported Node type: 27
Unsupported Node type: 27
Unsupported Node type: 27
Unsupported Node type: 27
Unsupported Node type: 27
Unsupported Node type: 27
Unsupported Node type: 27
Warning: block stack is not empty!
Unsupported Node type: 27
Unsupported Node type: 27
Warning: block stack is not empty!

```
attach file
[model.pyc.zip](https://github.com/zrax/pycdc/files/13222609/model.pyc.zip)

ddouworld:
Thank you very much for providing model.pyc, which will be very helpful to analyze why this error occurred. I'll try to run it. The class error is identified as a function when it appears to be decompiled.
![078 %5V44$~%PZFN56K0MRI](https://github.com/zrax/pycdc/assets/130207156/e64f74cd-a02d-4ecf-9ced-890bf706b1f3)

TiZCrocodile:
@zrax 
#458 fixes `Unsupported Node type: 27`
now it shows only this errors:
```
Unsupported argument found for LIST_EXTEND
Unsupported opcode: LIST_TO_TUPLE
Unsupported opcode: CALL_FUNCTION_EX
Unsupported opcode: DICT_MERGE
Unsupported argument found for LIST_EXTEND
Unsupported opcode: LIST_TO_TUPLE
Warning: block stack is not empty!
Warning: block stack is not empty!
```
this part decompiled with no errors now: (same part as the screenshot above)
![image](https://github.com/zrax/pycdc/assets/68035827/13f00454-83d8-4cf1-a3b1-c0e071d061b7)

so can close this also probably

PR #411, opened 2023-12-05
Add support for printing ASTNode::NODE_CLASS

greenozon:
This fixes the issue #408


greenozon:
it also helps to improve the output for the issue  #395

as follows;

before
```
def GEGLU():
    '''GEGLU'''

...
GEGLU = (GEGLU, 'GEGLU', nn.Module)

```

after
```
class GEGLU(nn.Module):
```

ASTree.cpp
```diff
@@ -3290,6 +3290,33 @@ void print_src(PycRef<ASTNode> node, PycModule* mod, std::ostream& pyc_output)
...
+            }
+            else {
+                pyc_output << ":\n";
+            }
+            printClassDocstring = true;
+            PycRef<ASTNode> code = node.cast<ASTClass>()->code().cast<ASTCall>()
+                ->func().cast<ASTFunction>()->code();
+            print_src(code, mod, pyc_output);
+        }
+        break;
```
zrax, line 3319: This looks like a duplicate of the code in `NODE_STORE` (starting on line 3146)...  It would be good to also replace that block with a call to `print_src()` to avoid code duplication.

zrax (changes requested):
Please also add a test for this case.

ASTree.cpp
greenozon: Well, I wish there would be just 1 piece of code for both cases, but after some debugging it turned out that those are different a bit..
eg: in original case with NODE_STORE there are 2 objects, one is ASTName that keeps the class name and other is the code itself
in this case the ASTName is absent, so I"m deducing the class name from the ASTClass object
Just added new commit for that missed printout of class name itself

2) I tried to create a test case, but to be honest I'm not sure how to create a sample one... when I try to define a class at root level its .pyc jumps  into old NODE_STORE case!
thus I need some help here please

Issue #387, opened 2023-10-02
Segmentation fault


b2nzy:
python 3.11 Segmentation fault

root@ruvds-sc7w2:~# ./pycdc edit.pyc
# Source Generated with Decompyle++
# File: edit.pyc (Python 3.11)

Segmentation fault

b2nzy:
[https://cdn.discordapp.com/attachments/1127852118846480484/1158422829700763648/edit.pyc?ex=651c30d5&is=651adf55&hm=a5de1f0f1554d556585134c46ef5363bc1a3b6d1069b9c752536d06475e71fb1&](url)

pyc file

greenozon:
@b2nzy 
"This content is no longer available."

PR #415, opened 2023-12-11
Gaurd against pop from empty block stack

dennis-doyensec:
I got a segfault when a block stack pop, when the block stack was already empty. This pull prevents emptying the block stack unless you are on the last instruction of the file. Or this is what I intended at least, I am new to this code base and pyc files in general.
 
This will prevent segfault seen in issue #387, I got a similar segfault independently. This pull prevents the segfault but the segfault would not occur if blocks were correctly parsed on a legitimate pyc file. So I don't know if this is really a fix for #387.

### DIFF
--- a/ASTree.cpp
+++ b/ASTree.cpp
@@ -1306,6 +1306,12 @@ PycRef<ASTNode> BuildFromCode(PycRef<PycCode> code, PycModule* mod)
                     stack_hist.pop();
                 }
 
+                if (blocks.size() == 1 && !source.atEof()) {
+                    fprintf(stderr, "Warning: Refusing to pop last block when there is more code to parse pos: %d OP: %02x\n", pos, opcode & 0xff);
+                    cleanBuild = false;
+                    break;
+                }
+
                 PycRef<ASTBlock> prev = curblock;
                 PycRef<ASTBlock> nil;
                 bool push = true;
@@ -1587,6 +1593,12 @@ PycRef<ASTNode> BuildFromCode(PycRef<PycCode> code, PycModule* mod)
                 if (curblock->nodes().size() &&
                         curblock->nodes().back().type() == ASTNode::NODE_KEYWORD) {
                     curblock->removeLast();
+                }
+
+                if (blocks.size() == 1 && !source.atEof()) {
+                    fprintf(stderr, "Warning: Refusing to pop last block when there is more code to parse pos: %d OP: %02x\n", pos, opcode & 0xff);
+                    cleanBuild = false;
+                    break;
                 }
 
                 if (curblock->blktype() == ASTBlock::BLK_IF


greenozon:
I guess one need to understand the root reason...
such kind of fix is a blurring real issue under the hood..
IMHO

dennis-doyensec:
Sorry for late response. 

In most cases I would agree with you 100%. I don't like half-fixes that hide
bugs. In this case, I think it depends on what you want this tool to do.

If you want pycdc to decompile legitimately created pyc files, stick with what
you have. Although, I would rather get an exit with a good warning, then a
segfault.

If you want this tool to also be useful for malware analysis, I would make a
change. Probably a better change then what I proposed. This bug can be used to
make a pyc file that can't be decompiled but can be loaded. 

The following pyc will run fine despite being invalid.
```
$ python3 -c 'import fun'      ## loads fun.pyc without any problems
Never seen

$ pycdc fun.pyc                ## pycdc with this patch (see warning)
# File: fun.pyc (Python 3.11)

Warning: Refusing to pop last block when there is more code to parse pos: 18 OP: 8c
if 1 == 4:
    pass
None()
print('Never seen')

$ pycdc fun.pyc                ## latest pycdc (2da061fc985) crashes, no useful output
# File: fun.pyc (Python 3.11)

[1]    13838 segmentation fault  pycdc fun.pyc

$ pycdas fun.pyc              ## assembly shows how the trick worked (invalid dead code)
fun.pyc (Python 3.11)
[Code]
    File Name: fun.py
    Object Name: <module>
    Qualified Name: <module>
    Arg Count: 0
    Pos Only Arg Count: 0
    KW Only Arg Count: 0
    Stack Size: 3
    Flags: 0x00000000
    [Names]
        'print'
    [Locals+Names]
    [Constants]
        1
        4
        'Never seen'
        None
    [Disassembly]
        0       RESUME                          0
        2       LOAD_CONST                      0: 1
        4       LOAD_CONST                      1: 4
        6       COMPARE_OP                      2 (==)
        12      POP_JUMP_FORWARD_IF_FALSE       10 (to 34)
        14      JUMP_FORWARD                    0 (to 16)
        16      JUMP_FORWARD                    0 (to 18)
        18      PRECALL                         0
        22      CALL                            0
        32      POP_TOP
        34      PUSH_NULL
        36      LOAD_NAME                       0: print
        38      LOAD_CONST                      2: 'Never seen'
        40      PRECALL                         1
        44      CALL                            1
        54      POP_TOP
        56      LOAD_CONST                      3: None
        58      RETURN_VALUE
```

murisfurder (approved):

greenozon:
@dennis-doyensec 
what was the original python code for the test case above?

at least two consecutive jumps looks very suspicious, isn't it?

```
        14      JUMP_FORWARD                    0 (to 16)
        16      JUMP_FORWARD                    0 (to 18)
```

I tried this
````
if 1 == 4:
    pass
print('Never seen')
```
no issues with both pycdas as well as pycdc

PR #424, opened 2024-01-05
LOAD_ATTR operand changes in 3.12

ncaklovic:
See changes in [LOAD_ATTR](https://docs.python.org/3.12/library/dis.html#opcode-LOAD_ATTR)  for 3.12

### DIFF
--- a/ASTree.cpp
+++ b/ASTree.cpp
@@ -1473,9 +1473,7 @@ PycRef<ASTNode> BuildFromCode(PycRef<PycCode> code, PycModule* mod)
             {
                 PycRef<ASTNode> name = stack.top();
                 if (name.type() != ASTNode::NODE_IMPORT) {
-                    auto arg = operand;
-                    if (mod->verCompare(3, 12) >= 0 && (arg & 1))
-                        arg >>= 1;
+                    auto arg = (mod->verCompare(3, 12) >= 0) ? operand >> 1 : operand;
                     stack.pop();
                     stack.push(new ASTBinary(name, new ASTName(code->getName(arg)), ASTBinary::BIN_ATTR));
                 }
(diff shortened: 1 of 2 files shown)


Mouad-scriptz (approved):

bytecode.cpp
```diff
@@ -390,7 +390,10 @@ void bc_disasm(std::ostream& pyc_output, PycRef<PycCode> code, PycModule* mod,
             case Pyc::LOAD_METHOD_A:
             case Pyc::LOAD_FROM_DICT_OR_GLOBALS_A:
                 try {
-                    formatted_print(pyc_output, "%d: %s", operand, code->getName(operand)->value());
+                    auto arg = operand;
+                    if (opcode == Pyc::LOAD_ATTR_A && mod->verCompare(3, 12) >= 0 && (arg & 1))
```
zrax, line 394:
```suggestion
                    if (opcode == Pyc::LOAD_ATTR_A && mod->verCompare(3, 12) >= 0)
```
The shift in Python 3.12+ is unconditional (the flag stored at the lowest order bit is used independently of the name lookup)

ASTree.cpp
```diff
@@ -1473,8 +1473,11 @@ PycRef<ASTNode> BuildFromCode(PycRef<PycCode> code, PycModule* mod)
             {
                 PycRef<ASTNode> name = stack.top();
                 if (name.type() != ASTNode::NODE_IMPORT) {
+                    auto arg = operand;
+                    if (mod->verCompare(3, 12) >= 0 && (arg & 1))
+                        arg >>= 1;
```
zrax, line 1478:
```suggestion
                    auto arg = (mod->verCompare(3, 12) >= 0) ? operand >> 1 : operand;
```

zrax (changes requested):

PR #437, opened 2024-02-13
Fixes #321 for arithmetic operators

### LINKED ISSUE #321: Operator precedence not being recovered by pycdc.
Decompiled output from `pycdc` doesn't follow the operator precedence from the original code.

Example:
ops.py
```py
def foo(a, b, c):
    return c - (a + b) 

def bar(a, b, c, d):
    return a / ((b ** c) * d)

def baz(a, b, c, d):
    return a - ((b ^ c) + d)
```

After compiling and decompiling, the decompiled output would look like this:
```py
# File: ops.cpython-310.pyc (Python 3.10)


def foo(a, b, c):
    return c - a + b


def bar(a, b, c, d):
    return a / b ** c * d


def baz(a, b, c, d):
    return a - (b ^ c) + d
```

The disassembly of the functions provide clarity to the precedence and the actual order of operations.
```
disassembly of foo:
  2           0 LOAD_FAST                2 (c)
              2 LOAD_FAST                0 (a)
              4 LOAD_FAST                1 (b)
              6 BINARY_ADD
              8 BINARY_SUBTRACT
             10 RETURN_VALUE

disassembly of bar:
  5           0 LOAD_FAST                0 (a)
              2 LOAD_FAST                1 (b)
              4 LOAD_FAST                2 (c)
              6 BINARY_POWER
              8 LOAD_FAST                3 (d)
             10 BINARY_MULTIPLY
             12 BINARY_TRUE_DIVIDE
             14 RETURN_VALUE

disassembly of baz:
  8           0 LOAD_FAST                0 (a)
              2 LOAD_FAST                1 (b)
              4 LOAD_FAST                2 (c)
              6 BINARY_XOR
              8 LOAD_FAST                3 (d)
             10 BINARY_ADD
             12 BINARY_SUBTRACT
             14 RETURN_VALUE
```

Maybe it would be nice to group each operation in parentheses?

Reference: [Operator Precedence from Python Documentation](https://docs.python.org/3/reference/expressions.html#operator-precedence)

kako57:
This fixes the operator precedence issue stated in #321, where the decompiler does not preserve the proper order of operations in the decompiled output.

This fix applies for arithmetic operations, specifically when the decompiler sees an add and subtract operation, or a divide and multiply at the same level of precedence.


zrax (approved):
I added a test case for it based on the referenced issue.  Otherwise, looks great, thanks!

Issue #443, opened 2024-02-19
Error decompyling lcu.pyc: vector::_M_range_check: __n (which is 62) >= this->size() (which is 61)


flyingwr:
[lcu.zip](https://github.com/zrax/pycdc/files/14330736/lcu.zip)

Python 3.12

greenozon:
not perfect, some try..
[lcu-issue443.cdc1.txt](https://github.com/zrax/pycdc/files/14334745/lcu-issue443.cdc1.txt)

zrax:
Should be fixed with #462

Issue #439, opened 2024-02-17
Error decompyling Serialchecker.pyc: vector::_M_range_check: __n (which is 24) >= this->size() (which is 22)


Myamii:
[Serialchecker.zip](https://github.com/zrax/pycdc/files/14319909/Serialchecker.zip)

i cant decompile it Error decompyling Serialchecker.pyc: vector::_M_range_check: __n (which is 24) >= this->size() (which is 22)

Python Version 3.12

greenozon:
some try, also following cases hit during process:

Warning: Stack history is not empty!
Warning: block stack is not empty!
Warning: block stack is not empty!
Warning: block stack is not empty!
Unsupported opcode: LOAD_SUPER_ATTR
Warning: Stack history is not empty!
Warning: block stack is not empty!
[Serialchecker.cdc1.txt](https://github.com/zrax/pycdc/files/14322067/Serialchecker.cdc1.txt)

Myamii:
> some try, also following cases hit during process:
> 
> Warning: Stack history is not empty! Warning: block stack is not empty! Warning: block stack is not empty! Warning: block stack is not empty! Unsupported opcode: LOAD_SUPER_ATTR Warning: Stack history is not empty! Warning: block stack is not empty! [Serialchecker.cdc1.txt](https://github.com/zrax/pycdc/files/14322067/Serialchecker.cdc1.txt)

thanks!

zrax:
Should be fixed with #462

Issue #430, opened 2024-02-05
Error decompyling deobs_mine.pyc: vector::_M_range_check: __n (which is 10) >= this->size() (which is 7)


shoaibhassan0:
[deobs_mine.zip](https://github.com/zrax/pycdc/files/14165530/deobs_mine.zip)
i am also attaching this pyc file 
i am using command pycdc -c -v 3.11 deobs_mine.pyc 
on windows 10
When i was decompiling the python marshal object bytecode it results this

E:\build>pycdc -c -v 3.11 deobs_mine.pyc
# Source Generated with Decompyle++
# File: deobs_mine.pyc (Python 3.11)

import requests
import json
import time
import sys
from platform import system
import os
import subprocess
import http.server as http
import socketserver
import threading
from datetime import datetime
import uuid
[...]

Issue #418, opened 2023-12-16
Error decompyling {file directory}: vector


anandh8x:
![Screenshot_20231216-215957](https://github.com/zrax/pycdc/assets/128164409/aedfa92d-f0b8-453d-903a-0d728762f5fa)

I have three files which I want to convert to py 
But getting error like shown above

greenozon:
Not enough data to debug by your picture only

anandh8x:
[a.zip](https://github.com/zrax/pycdc/files/13709578/a.zip)
Here's the file @greenozon

greenozon:
@anandh8x 
main issue (multiple cases)
is as follows:

```
Unsupported opcode: RETURN_GENERATOR
```

anandh8x:
There is no solution for it ?

greenozon:
wow, there are much much more unsupported opcodes...
(besides above noted)

greenozon:
there are still many async/with/etc issues but you could get the high level overview of the code

[a.cdc1.zip](https://github.com/zrax/pycdc/files/13720471/a.cdc1.zip)

zrax:
Should be fixed with #462

PR #462, opened 2024-02-27
Aligning some opcodes for Python 3.11, 3.12: LOAD_GLOBAL, LOAD_ATTR

greenozon:
According to official documentation
https://docs.python.org/3/library/dis.html#opcode-LOAD_GLOBAL
https://docs.python.org/3/library/dis.html#opcode-LOAD_ATTR

### DIFF
--- a/ASTree.cpp
+++ b/ASTree.cpp
@@ -1480,6 +1480,18 @@ PycRef<ASTNode> BuildFromCode(PycRef<PycCode> code, PycModule* mod)
                 PycRef<ASTNode> name = stack.top();
                 if (name.type() != ASTNode::NODE_IMPORT) {
                     stack.pop();
+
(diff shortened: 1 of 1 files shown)


greenozon:
#439   looks like similar root cause, exception is gone

greenozon:
probably #356   (no input provided, but error output and python version 3.11 matches I guess..)

TiZCrocodile:
yea you right, its my bad I added the support for `LOAD_GLOBAL`, but the low bit of operand indicates to push null, but always `operand >>= 1` needs to happen if its 3.11+, and not only when low bit is set

TiZCrocodile:
this closes all this issues about vector range in python 3.11
#443 #439 #430 #418 #407 #403 #398 #389 #379 #356

im really sorry, its my fault

zrax (approved):

Issue #304, opened 2022-12-15
make[2]: *** [CMakeFiles/pycxx.dir/bytecode.cpp.o] Error 1


BaseMax:
```
[root@5-161-122-190 build]# yum install cmake
Loaded plugins: fastestmirror, universal-hooks
Loading mirror speeds from cached hostfile
 * EA4: 63.247.64.58
 * cpanel-plugins: 63.247.64.58
 * base: mirror.arizona.edu
 * epel: fedora.westmancom.com
 * extras: centos.mirrors.hoobly.com
 * updates: centos.hivelocity.net
Mysql-connectors-community                                                                 | 2.6 kB  00:00:00
Mysql-tools-community                                                                      | 2.6 kB  00:00:00
Mysql57-community                                                                          | 2.6 kB  00:00:00
base                                                                                       | 3.6 kB  00:00:00
extras                                                                                     | 2.9 kB  00:00:00
imunify360                                                                                 | 3.0 kB  00:00:00
imunify360-rollout-1                                                                       | 3.0 kB  00:00:00
imunify360-rollout-2                                                                       | 3.0 kB  00:00:00
imunify360-rollout-3                                                                       | 3.0 kB  00:00:00
imunify360-rollout-4                                                                       | 3.0 kB  00:00:00
updates                                                                                    | 2.9 kB  00:00:00
(1/2): imunify360/7/primary_db                                                             | 1.2 MB  00:00:00
(2/2): updates/7/x86_64/primary_db                                                         |  18 MB  00:00:00
Resolving Dependencies
--> Running transaction check
---> Package cmake.x86_64 0:2.8.12.2-2.el7 will be installed
--> Processing Dependency: libarchive.so.13()(64bit) for package: cmake-2.8.12.2-2.el7.x86_64
--> Running transaction check
---> Package libarchive.x86_64 0:3.1.2-14.el7_7 will be installed
--> Finished Dependency Resolution

Dependencies Resolved

==================================================================================================================
 Package                     Arch                    Version                          Repository             Size
==================================================================================================================
Installing:
 cmake                       x86_64                  2.8.12.2-2.el7                   base                  7.1 M
Installing for dependencies:
 libarchive                  x86_64                  3.1.2-14.el7_7                   base                  319 k

Transaction Summary
==================================================================================================================
Install  1 Package (+1 Dependent package)

Total download size: 7.4 M
Installed size: 27 M
Is this ok [y/d/N]: y
Downloading packages:
[...]

PR #469, opened 2024-03-08
Fix for Centos6/7 compilation issues

greenozon:
Centos6/7 uses very old gcc/g++ so some alignments should be done in order to support it
those are minor AFAIK
Fixes issue https://github.com/zrax/pycdc/issues/304

After this PR the output on Centos7 is as follows:

image

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,7 +17,7 @@ if (ENABLE_STACK_DEBUG)
 endif()
 
 if(CMAKE_COMPILER_IS_GNUCXX OR "${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
-    set(CMAKE_CXX_FLAGS "-Wall -Wextra -Wshadow -Werror ${CMAKE_CXX_FLAGS}")
+    set(CMAKE_CXX_FLAGS "-Wall -Wextra -Wno-error=shadow -Werror ${CMAKE_CXX_FLAGS}")
 elseif(MSVC)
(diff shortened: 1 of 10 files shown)


CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ if (ENABLE_STACK_DEBUG)
 endif()
 
 if(CMAKE_COMPILER_IS_GNUCXX OR "${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
-    set(CMAKE_CXX_FLAGS "-Wall -Wextra -Wshadow -Werror ${CMAKE_CXX_FLAGS}")
+    set(CMAKE_CXX_FLAGS "-Wall -Wextra -Werror ${CMAKE_CXX_FLAGS}")
```
zrax, line 20: This will disable all types of shadowing warnings, not just the overzealous one that GCC 4.8 had...  I would prefer the parameters be renamed and this warning kept to make GCC happy.

zrax (changes requested):

CMakeLists.txt
greenozon: makes sense!
agree, but would it be fine with you if I rename all those shadowed input params (mostly named as `type` into `_type`) ?

zrax: Yeah, that would be fine

zrax:
Well, there certainly are a lot more of those than I expected...  I think I've changed my mind, and would prefer `-Wno-error=shadow`, so they still show up as warnings, but won't block compilation.

greenozon:
@zrax   addressed your request

zrax (approved):

PR #485, opened 2024-05-27
Added support for SWAP and WITH_EXCEPT_START

ddouworld:
WITH_EXCEPT_START is simply put together with SETUP_WITH_A, but it works fine.
The test code for SWAP is as follows
```
def SWAP():
    my_array = [
        1,
        2,
        3,
        4,
        5,
        6,
        8]
    i = 1
    j = 3
    my_array[i], my_array[j], my_array[2] = my_array[j], my_array[i], my_array[4]
```
The generated code looks like this
```
def SWAP():
    my_array = [
        1,
        2,
        3,
        4,
        5,
        6,
        8]
    i = 1
    j = 3
    (my_array[i], my_array[j], my_array[2]) = (my_array[j], my_array[i], my_array[4])
```
It's not quite the same, but I don't think it should affect function much


TiZCrocodile:
the problem with this:
```
my_array[i] = my_array[j]
my_array[j] = my_array[i]
my_array[2] = my_array[4]
```
is that the logic changes, for example
```
x=1
y=2
x,y = y,x
```
would replace them, but this
 ```
 x=1
y=2
x = y
y = x
 ```
would result in both variables equals to y, because after x = y, now y = x, means y = y

ddouworld:
Yes, you're right. I think I need to change it

PR #490, opened 2024-06-11
Run tests in parallel or only some with a filter

Levak:
I rewrote the `all_tests.sh` file to spread all tests among a set amount of workers so that it runs faster on machines with lots of cores. And since I was also trying to make some fixes in `pycdc` I needed to run only some tests of the fixtures I was touching.

`make check JOBS=4` will spawn 4 workers and spread the tests among them.
`make check FILTER=try` will only run tests that contain `try` in their name.


TiZCrocodile:
I think its may be good to have the option to test certain tests, but I also think that after every change, all tests should be checked, because sometimes things rely on each other and you may broke other things accidentally

tests/all_tests.sh
```diff
@@ -1,13 +1,9 @@
...
-for tf in "${test_files[@]}"; do
-    test_name="$(basename "$tf")"
-    test_name="${test_name%.txt}"
-    "$srcdir"/tests/decompyle_test.sh $test_name tests || test_status=1
-done
-
-exit $test_status
+find tests/tokenized -type f -name '*.txt' -a -name "*${filter}*" -print0 | \
+    xargs -0 -I '{}' -P $jobs \
+          bash -c 'echo -e "$(tests/decompyle_test.sh "$(basename -s .txt "{}")" tests-out)"'
```
zrax, line 9: This no longer fails when the tests fail...  The exit status must be preserved and propagated (perhaps just a `set -e` is sufficient?)

tests/all_tests.sh
```diff
@@ -1,13 +1,9 @@
 #!/bin/bash
 
-srcdir="$(cd "$(dirname "${BASH_SOURCE[0]}")"/.. && pwd)"
+cd "$(dirname "$0")"/..
```
zrax, line 3: The original script didn't change directory because the output directory, as well as the path to the generated pycdc executable, was relative to the build path...  This version of the script no longer works with out-of-tree builds.

tests/all_tests.sh
```diff
@@ -1,13 +1,9 @@
...
-test_status=0
-test_files=( "$srcdir"/tests/tokenized/*.txt )
-for tf in "${test_files[@]}"; do
-    test_name="$(basename "$tf")"
-    test_name="${test_name%.txt}"
-    "$srcdir"/tests/decompyle_test.sh $test_name tests || test_status=1
-done
-
-exit $test_status
+find tests/tokenized -type f -name '*.txt' -a -name "*${filter}*" -print0 | \
```
zrax, line 7: Minor nitpick, but it would be nice to only add the `-a -name "*${filter}*"` parameters when FILTER is specified.

zrax (changes requested):

tests/all_tests.sh
Levak: I'm not sure what you refer to as "out-of-tree builds". 

`$0` refers to the program name called from the command line, e.g. `./tests/all-test.sh` or `/absolute/path/tests/all-tests.sh`
`dirname $0` will get the parent directory, which we use as a base to change current working directory. The change of directory is only effective inside the subshell of the script (and all its children). The underlying script `decompyle_test.sh` will use relative paths from that.

Maybe I am missing one scenario where this fails, are you able to develop on this?

Levak:
> I think its may be good to have the option to test certain tests, but I also think that after every change, all tests should be checked, because sometimes things rely on each other and you may broke other things accidentally

Yes, I totally agree with you on "a change in the code should be tested against all tests", but from my POV, in this project, when you're trying to implement one opcode, it rarely breaks others, besides, you just want to quickly check if it works with the intended test, and _then_ run all tests to make sure you didn't break everything.

tests/all_tests.sh
Levak: I think it adds too much complexity when this one-line works. When FILTER is not defined or empty, the globing will become `**` which will match everything. My goal with this PR was to add simple parallelism without trying to re-implement pytest-xdist in bash ^^.

tests/all_tests.sh
Levak: I also had to rework the inner subshell:
```diff
- bash -c 'echo -e "$(tests/decompyle_test.sh "$(basename -s .txt "{}")" tests-out)"'
+ bash -c 'o=$(tests/decompyle_test.sh "$(basename -s .txt "{}")" tests-out) r=$?; echo "$o"; exit $r'

```

tests/all_tests.sh
zrax: out-of-tree builds mean when the build tree is not directly in the source tree (which is generally the preferred way of using cmake).  For example, see the GitHub workflows, which do the build in a /build directory.

This script doesn't directly deal with the path to the built pycdc executable, but the called script (decompyle_test.sh) does, and it currently assumes that pycdc is in the cwd.

Levak: Alright, thanks for taking the time to explain the issue - I do see it now and it is a terrible mistake. I've opted for a similar approach that was in the original script.
I also chose to squash all the commits as they were part of the review process.
Cheers

Levak: I think I now understand why you use this construction:
```
srcdir="$(cd "$(dirname "${BASH_SOURCE[0]}")"/.. && pwd)"
```
Instead of a simple
```
srcdir="$(dirname "$0")"
```
_Thank you Windows back slashes_
Fix on the way.

zrax (approved):

PR #492, opened 2024-06-21
Fix Long Numeric Integer representation for Python 3

Levak:
Before Python 3, long integers were input with an `L` suffix. Since Python 3, all integers are 64-bit and do not need the `L` suffix.
Source: https://docs.python.org/3.0/whatsnew/3.0.html#integers

Scripts generated by `pycdc` from Python 3 that contained long numeric integers would not work on Python 3 because it always suffixed them with `L`.

Note: I am not sure how to adapt the test suite to reflect this.
Should I just add the `.pyc` file without changing `test_integers.py` ? That file was made with Python 2 in mind (e.g. `print` without parenthesis).
EDIT: Nvm, there is too many differences (e.g. `sys.maxint` was removed in Python 3).

### DIFF
--- a/bytecode.cpp
+++ b/bytecode.cpp
@@ -230,7 +230,7 @@ void print_const(std::ostream& pyc_output, PycRef<PycObject> obj, PycModule* mod
         formatted_print(pyc_output, "%d", obj.cast<PycInt>()->value());
         break;
     case PycObject::TYPE_LONG:
-        formatted_print(pyc_output, "%s", obj.cast<PycLong>()->repr().c_str());
+        formatted_print(pyc_output, "%s", obj.cast<PycLong>()->repr(mod).c_str());
         break;
     case PycObject::TYPE_FLOAT:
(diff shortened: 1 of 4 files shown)


Levak:
Maybe we can change `test_integers.py` this way?

# Python 3
```
$ echo 'print(24**24)' | python3 -m dis -
  1           0 LOAD_NAME                0 (print)
              2 LOAD_CONST               2 (1333735776850284124449081472843776)
              4 CALL_FUNCTION            1
              6 POP_TOP
              8 LOAD_CONST               1 (None)
             10 RETURN_VALUE
```

# Python 2
```
$ echo 'print(24**24)' | python2.7 -m dis -
  1           0 LOAD_CONST               2 (1333735776850284124449081472843776L)
              3 PRINT_ITEM          
              4 PRINT_NEWLINE       
              5 LOAD_CONST               1 (None)
              8 RETURN_VALUE        
```

zrax (approved):

PR #493, opened 2024-06-24
Implement Python 3.8 && 3.9 try-except-finally logic

Levak:
- Add tests for simple cases of try-except-finally (rename one old Python 2.6 test);

- Implement `JUMP_IF_NOT_EXC_MATCH_A` with a fake `CMP_EXCEPTION` followed by a `POP_JUMP_IF_TRUE`;

- Implement `BEGIN_FINALLY` with a fake `LOAD_CONST_A None`;

- Implement Python 3.8+ specific `SETUP_FINALLY_A` handling both EXCEPT and FINALLY cases. We guess which is which by looking at the targetted block that seems to either be a POP_TOP or a DUP_TOP in the case of an EXCEPT block;

- Implement RERAISE with a fake END_FINALLY;

- In Python 3.9+, skip the duplicated code from the FINALLY blocks with a jump in bytecode and a fake `END_FINALLY`.


EchoStretch:
Fantastic update! I think this one works the best so far. There are still a few I'm not able to resolve, but I'm reviewing everyone’s repos to get more insights.

Unsupported opcode: LOAD_ASSERTION_ERROR
Unsupported opcode: LOAD_ASSERTION_ERROR
Unsupported opcode: LOAD_ASSERTION_ERROR
Warning: Stack history is not empty!
Warning: block stack is not empty!
Warning: block stack is not empty!
Warning: block stack is not empty!
Warning: Stack history is not empty!
Warning: block stack is not empty!
Unsupported opcode: LOAD_ASSERTION_ERROR
Unsupported opcode: LOAD_ASSERTION_ERROR
Warning: block stack is not empty!
Unsupported opcode: LOAD_ASSERTION_ERROR
Unsupported opcode: LOAD_ASSERTION_ERROR

World-sudo:
was it merged?

plusls:
any progress?

Levak:
> any progress?

No, I gave up on the project when I realized that new Python bytecode requires context checking trees rather than per-bytecode translation. Newer Python like 3.8+ optimize more and more the bytecode to remove unnecessary instructions for the execution which inherently destroys the one-to-one bijection that was before, where one instruction could result in only one step/line of python code . 

I used this tool and my fork to do the heavy lifting of decompiling a python 3.9 program I needed the sources of to run on linux, and when it would fail to decompile, I would decompile by hand the bytecode.  I managed to decompile what I needed then I moved to other things. Seeing that the project is basically too anchored in previous versions of Python, it needs a complete rewrite for python 3.9+ I simply stopped all work. 

I haven't looked ever since if there was any better suited decompiler for python 3.8+ since I left this PR.

PR #511, opened 2024-08-13
fix: incorrect error message in `ASTree::BuildFromCode(...)`

wilson0x4d:
* `opcode` was incorrectly clamped causing misleading output.

* modified output with additional info about offending bytecode and position so end-user submissions can become more informative.

### DIFF
--- a/ASTree.cpp
+++ b/ASTree.cpp
@@ -2474,7 +2474,8 @@ PycRef<ASTNode> BuildFromCode(PycRef<PycCode> code, PycModule* mod)
             }
             break;
         default:
-            fprintf(stderr, "Unsupported opcode: %s\n", Pyc::OpcodeName(opcode & 0xFF));
+            int bytecode = code->code()->value()[curpos] & 0xFF;
+            fprintf(stderr, "Unsupported opcode: %s (bytecode 0x%02x (%d) at position %d)\n", Pyc::OpcodeName(opcode), bytecode, bytecode, curpos);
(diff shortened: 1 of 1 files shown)


ASTree.cpp
```diff
@@ -2474,7 +2474,8 @@ PycRef<ASTNode> BuildFromCode(PycRef<PycCode> code, PycModule* mod)
             }
             break;
         default:
-            fprintf(stderr, "Unsupported opcode: %s\n", Pyc::OpcodeName(opcode & 0xFF));
+            int bytecode = code->code()->value()[curpos] & 0xFF;
```
zrax, line 2477: Reading directly from the code object doesn't handle the `EXTENDED_ARG` case properly...  Perhaps just printing the position is sufficient, since it can be used to look up the context in disassembly.

wilson0x4d:
indeed, i overlooked the `EXTENDED_ARG` handling.

the update i've pushed instead relies on `bc_next` to pass `bytecode` back up to the caller, this seems cleaner (doesn't couple the code to `code()->value()`) and compensates for `EXTENDED_ARG` correctly. this does leave the position printed off by a few bytes, but the bytecode value becomes accurate (which is what I was really looking for in the output) and even being a few bytes off at least gets someone who understands the bytecode into the right area of the image (where they would see `EXTENDED_ARG` and could mentally compensate.)

i also updated the disasm code with a similar change (to include the offending bytecode in the output.)

Issue #507, opened 2024-08-07
Wrong default function parameters decompilation


greenozon:
input sample:

```
def foo1(p1, p2, /, arg1=100, arg2='argos2', **kwargs):
	pass
```

decompiled using pycdc as:

```
def foo1(p1, p2, arg1, arg2 = (100, 'argos2'), **kwargs):
    pass
```

any clues how to fix it to the rescue?

sample files attached

[sample-1.zip](https://github.com/user-attachments/files/16533453/sample-1.zip)

Issue #448, opened 2024-02-21
Support Python 3.6 decompilation


zrax:
Tasks
====
- [X] Bytecode support (pycdas)
- [ ] Handle new opcodes in AST builder
  - [X] `SETUP_ANNOTATIONS`
  - [ ] `STORE_ANNOTATION`
  - [ ] `CALL_FUNCTION_EX`
  - [X] `FORMAT_VALUE`
  - [X] `BUILD_CONST_KEY_MAP`
  - [X] `BUILD_STRING`
  - [ ] `BUILD_TUPLE_UNPACK_WITH_CALL`
- [ ] Handle changes in `CALL_FUNCTION_*` / `MAKE_FUNCTION` operands

PR #539, opened 2025-02-13
Fix MAKE_FUNCTION default args for py 3.7

CarrBen:
DRAFT: This is stacked on top of #538 as both include changes to the same opcode

Some changes for the handling of MAKE_FUNCTION operand & arguments in py3.6-3.12 that I discovered while make some fixes to decompile a py 3.11 app. I started on this before I noticed that #391 existed, but as this PR is stacked on top of #538 it works for py 3.0 & py 3.4 too.

[MAKE_FUNCTION Python 3.4 Docs](https://docs.python.org/3.4/library/dis.html#opcode-MAKE_FUNCTION)
[MAKE_FUNCTION Python 3.7 Docs](https://docs.python.org/3.7/library/dis.html#opcode-MAKE_FUNCTION)

## Summary
* Remove xfail test_functions_py3 for py 3.7
* Update compiled test_functions_py3 for py 3.7 to match input .py with the new test case in #538 
* Added a branch for py 3.6+ to handle the args & kwargs being in tuples & dicts

### Tests
The `test_functions_py3` for py 3.7 was in xfail.
In the previous PR, I added a new test case and updated the tokenized version, to match this the `test_functions_py3.3.7.pyc` file in `compiled` was re-built from the (previously updated) source in `input`.
I used the `token_dump` and `pymultic` scripts for this, with a slight modification to use `-alpine` containers in `pymultic`.

### New py3.6-12 Arg Defaults for MAKE_FUNCTION
With the current `master`, `pycdc` produced this output for the `x7d` test case on py3.7.
```python
def x7d(foo = None, *, bar, **kwargs):
``` 
However, this is the input
```python
def x7d(foo = 1, *, bar = 2, **kwargs):
```
This is because the MAKE_FUNCTION code expects individual args on the stack, but they are now on the stack inside of 0-4 tuples & kwargs indicated by the operand flag.

Stacked on top of #538 
Duplicates #391 
Related Issues
#507 ?
#155 ?
#138 
#448 The MAKE_FUNCTION part


Luiz-Monad:
And they changed it again on Python 3.13

1111qqjwbbw:
Hello @CarrBen thanks for your contribution.You can update this PR to Python3.13 version 3.13?thank you!

CarrBen:
@whoami730 @1111qqjwbbw I think this branch might work OK for 3.12 (though there might be some other things that don't play nice with it).

For 3.13, I'm looking at updating it, I think I have a fix, but it looks like there are some other changes to CALL in 3.13 that have some wider consequences.

I might raise another PR for 3.13 support rather than push here, but I'll update here with more info once I push something.

CarrBen:
@whoami730 @1111qqjwbbw I've set up #579 for the Python 3.13 MAKE_FUNCTION. It's a draft as it builds on top of both this PR and #538

PR #543, opened 2025-03-20
Improved

Aasyaco:


zrax:
I don't know what you're trying to do here, but throwing a bunch of random opcodes into the jump-handling block is probably not very useful...

Aasyaco:
> I don't know what you're trying to do here, but throwing a bunch of random opcodes into the jump-handling block is probably not very useful...

Fixed reraise and added some opcodes

PR #472, opened 2024-03-16
Support for JUMP_BACKWARD

greenozon:
According to this article the JUMP_BACKWARD  was a replacement for the same opcode JUMP_ABSOLUTE
https://bugs.python.org/issue47120

**Test Case**
```
def JUMP_BACKWARD():
    iterable = [1, 2, 3]
    for item in iterable:
        pass

    while True:
        break
```

pycdas for TC
```
[Code]
    File Name: jmp_back.py
    Object Name: <module>
    Qualified Name: <module>
    Arg Count: 0
    Pos Only Arg Count: 0
    KW Only Arg Count: 0
    Stack Size: 1
    Flags: 0x00000000
    [Names]
        'JUMP_BACKWARD'
    [Locals+Names]
    [Constants]
        [Code]
            File Name: jmp_back.py
            Object Name: JUMP_BACKWARD
            Qualified Name: JUMP_BACKWARD
            Arg Count: 0
            Pos Only Arg Count: 0
            KW Only Arg Count: 0
            Stack Size: 2
            Flags: 0x00000003 (CO_OPTIMIZED | CO_NEWLOCALS)
            [Names]
            [Locals+Names]
                'iterable'
                'item'
            [Constants]
                None
                (
                    1
                    2
                    3
                )
            [Disassembly]
                0       RESUME                          0
                2       BUILD_LIST                      0
                4       LOAD_CONST                      1: (1, 2, 3)
                6       LIST_EXTEND                     1
                8       STORE_FAST                      0: iterable
                10      LOAD_FAST                       0: iterable
                12      GET_ITER
                14      FOR_ITER                        2 (to 20)
                18      STORE_FAST                      1: item
                20      JUMP_BACKWARD                   4 (to 14)
                22      END_FOR
                24      NOP
                26      RETURN_CONST                    0: None
        None
    [Disassembly]
        0       RESUME                          0
        2       LOAD_CONST                      0: <CODE> JUMP_BACKWARD
        4       MAKE_FUNCTION                   0
        6       STORE_NAME                      0: JUMP_BACKWARD
        8       RETURN_CONST                    1: None
```

**before**
```
def JUMP_BACKWARD():
Unsupported opcode: JUMP_BACKWARD
    iterable = [
        1,
        2,
        3]
# WARNING: Decompyle incomplete
```

**after**
```
def JUMP_BACKWARD():
Unsupported opcode: END_FOR
    iterable = [
        1,
        2,
        3]
# WARNING: Decompyle incomplete
```

PR #555, opened 2025-06-30
Support py 3.12 opcodes: part 1

whoami730:
- Added support for `END_FOR` and `JUMP_BACKWARD` opcodes for Python 3.12
- Added loop tests


whoami730:
This PR is based upon #472 for `JUMP_BACKWARD` support.

ASTree.cpp
```diff
@@ -897,7 +897,8 @@ PycRef<ASTNode> BuildFromCode(PycRef<PycCode> code, PycModule* mod)
         case Pyc::INSTRUMENTED_FOR_ITER_A:
             {
                 PycRef<ASTNode> iter = stack.top(); // Iterable
-                stack.pop();
+                if (mod->verCompare(3,12) < 0)
```
whoami730, line 900: This behaviour is changed in py 3.12. Need to add a comment for the same.

ASTree.cpp
```diff
@@ -1678,10 +1682,38 @@ PycRef<ASTNode> BuildFromCode(PycRef<PycCode> code, PycModule* mod)
...
+                if (curblock->blktype() == ASTBlock::BLK_FOR) {
+                    PycRef<ASTBlock> prev = blocks.top();
+                    blocks.pop();
+
+                    curblock = blocks.top();
+                    curblock->append(prev.cast<ASTNode>());
+                }
+                else {
+                    fprintf(stderr, "Wrong block type %i for END_FOR\n", curblock->blktype());
+                    break;
```
whoami730, line 1708: `break` not needed here

PR #557, opened 2025-06-30
Support py 3.12 opcodes: part 2

whoami730:
- Added support for `COPY`, `BINARY_SLICE` and `STORE_SLICE` opcodes for Python 3.12
- Added tests for `BINARY_SLICE` and `STORE_SLICE` opcodes, unable to find useful test for `COPY` opcode


ASTree.cpp
```diff
@@ -1678,10 +1682,38 @@ PycRef<ASTNode> BuildFromCode(PycRef<PycCode> code, PycModule* mod)
         case Pyc::POP_EXCEPT:
             /* Do nothing. */
             break;
+        case Pyc::END_FOR:
+            {
+                stack.pop();
+
+                if ((opcode == Pyc::END_FOR) && (mod->majorVer() == 3) && (mod->minorVer() == 12)) {
```
zrax, line 1689: Is this not needed for 3.13 and later?

whoami730:
Unable to find any code to generate `COPY` operand so far.

PR #564, opened 2025-07-15
Parse and display exception table

whoami730:
Parses and displays Exception Table starting Py 3.11+
This is required for supporting new "zero-cost" exception handling mechanism.

### DIFF
--- a/bytecode.cpp
+++ b/bytecode.cpp
@@ -600,3 +600,18 @@ void bc_disasm(std::ostream& pyc_output, PycRef<PycCode> code, PycModule* mod,
         pyc_output << "\n";
     }
 }
+
+void bc_exceptiontable(std::ostream& pyc_output, PycRef<PycCode> code,
+               int indent)
+{
+    for (auto tuple: code->exceptTableEntries()) {
+
+        for (int i=0; i<indent; i++)
+            pyc_output << "    ";
+
+        pyc_output << std::get<0>(tuple) << " to " << std::get<1>(tuple);
+        pyc_output << " -> " << std::get<2>(tuple) << " ";
+        pyc_output << "[" << std::get<3>(tuple) << "] " << (std::get<4>(tuple) ? "lasti": "");
+        pyc_output << "\n";
+    }
+}
--- a/pyc_code.cpp
+++ b/pyc_code.cpp
@@ -128,3 +128,44 @@ PycRef<PycString> PycCode::getCellVar(PycModule* mod, int idx) const
         ? m_freeVars->get(idx - m_cellVars->size()).cast<PycString>()
         : m_cellVars->get(idx).cast<PycString>();
 }
+
+int _parse_varint(PycBuffer& data, int& pos) {
+    int b = data.getByte();
+    pos += 1;
+
+    int val = b & 63;
+    while (b & 64) {
+        val <<= 6;
+
+        b = data.getByte();
+        pos += 1;
+
+        val |= (b & 63);
+    }
+    return val;
+}
+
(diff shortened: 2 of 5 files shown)


pyc_code.h
```diff
@@ -87,6 +87,10 @@ class PycCode : public PycObject {
         m_globalsUsed.emplace_back(std::move(varname));
     }
 
+    typedef std::tuple<int, int, int, int, bool> exception_table_entry_t;
```
zrax: I would prefer to see an actual struct here; otherwise the elements have no obvious meaning...

bytecode.cpp
```diff
@@ -600,3 +600,18 @@ void bc_disasm(std::ostream& pyc_output, PycRef<PycCode> code, PycModule* mod,
         pyc_output << "\n";
     }
 }
+
+void bc_exceptiontable(std::ostream& pyc_output, PycRef<PycCode> code,
+               int indent)
+{
+    for (auto tuple: code->exceptTableEntries()) {
```
zrax:
```suggestion
    for (const auto& tuple : code->exceptTableEntries()) {
```

bytecode.cpp
```diff
@@ -600,3 +600,18 @@ void bc_disasm(std::ostream& pyc_output, PycRef<PycCode> code, PycModule* mod,
...
+
+        for (int i=0; i<indent; i++)
+            pyc_output << "    ";
+
+        pyc_output << std::get<0>(tuple) << " to " << std::get<1>(tuple);
+        pyc_output << " -> " << std::get<2>(tuple) << " ";
+        pyc_output << "[" << std::get<3>(tuple) << "] " << (std::get<4>(tuple) ? "lasti": "");
+        pyc_output << "\n";
```
zrax:
```suggestion
        pyc_output << std::get<0>(tuple) << " to " << std::get<1>(tuple)
                   << " -> " << std::get<2>(tuple) << " "
                   << "[" << std::get<3>(tuple) << "] " << (std::get<4>(tuple) ? "lasti": "")
                   << "\n";
```

bytecode.cpp
```diff
@@ -600,3 +600,18 @@ void bc_disasm(std::ostream& pyc_output, PycRef<PycCode> code, PycModule* mod,
...
+    }
+}
```
zrax: Fix newline at EOF

pyc_code.cpp
```diff
@@ -128,3 +128,44 @@ PycRef<PycString> PycCode::getCellVar(PycModule* mod, int idx) const
...
+
+    int val = b & 63;
+    while (b & 64) {
+        val <<= 6;
+
+        b = data.getByte();
+        pos += 1;
+
+        val |= (b & 63);
+    }
```
zrax: Purely personal preference, but when bits and shifts are involved, I prefer to use hex instead of decimal.
```suggestion
    int val = b & 0x3F;
    while (b & 0x40) {
        val <<= 6;

        b = data.getByte();
        pos += 1;

        val |= (b & 0x3F);
    }
```

pyc_code.cpp
```diff
@@ -128,3 +128,44 @@ PycRef<PycString> PycCode::getCellVar(PycModule* mod, int idx) const
...
+        int dl = _parse_varint(data, pos);
+
+        int depth = dl >> 1;
+        bool lasti = bool(dl & 1);
+        
+        entries.emplace_back(start, end, target, depth, lasti);
+    }
+    
+    return entries;
+}
```
zrax: Newline at EOF

whoami730:
Thanks for the comments @zrax , I have updated the PR as per your suggestions.

PR #576, opened 2025-09-01
Parse cell & free vars for Py 3.11+

whoami730:
- Parses cell & free vars from `locals+names` for Py 3.11+
- Support new opcode `COPY_FREE_VARS` and `MAKE_CELL` with tests


pyc_code.cpp
```diff
@@ -26,6 +26,10 @@ lntable                         Obj     Obj     Obj     Obj     Obj     Obj
 exceptiontable                                                          Obj
 */
 
+#define CO_FAST_LOCAL 0x20
+#define CO_FAST_CELL 0x40
+#define CO_FAST_FREE 0x80
```
whoami730: Need to convert to enum

pyc_code.cpp
```diff
@@ -75,23 +79,49 @@ void PycCode::load(PycData* stream, PycModule* mod)
...
+                free_vars_extra.push_back(name);
+            }
+        }
+    }
     else
         m_localKinds = new PycString;
 
     if (mod->verCompare(2, 1) >= 0 && mod->verCompare(3, 11) < 0)
         m_freeVars = LoadObject(stream, mod).cast<PycSequence>();
+    else if (mod->verCompare(3, 11) >= 0)
```
whoami730, line 113: I think given the complicated structure this whole function should be rewritten appropriately.

ASTree.cpp
```diff
@@ -3567,6 +3580,21 @@ void decompyle(PycRef<PycCode> code, PycModule* mod, std::ostream& pyc_output)
             }
             pyc_output << "\n";
         }
+
+        PycCode::nonlocals_t non_locals = code->getNonLocals();
```
whoami730, line 3584: Easier to duplicate from globals instead of handling both together. Can be changed if need be.

pyc_code.cpp
whoami730: Honestly seems quite hard to rewrite this function. I think default initializations would be quite useful here.

whoami730: Default initializations themselves are not trivial :| It's annoying as well. I shall skip simplifying this logic at the moment.

whoami730:
Requesting review @greenozon @zrax

pyc_code.cpp
```diff
@@ -75,23 +76,49 @@ void PycCode::load(PycData* stream, PycModule* mod)
     m_consts = LoadObject(stream, mod).cast<PycSequence>();
     m_names = LoadObject(stream, mod).cast<PycSequence>();
 
+    // Represents localsplusnames for py 3.11+
     if (mod->verCompare(1, 3) >= 0)
```
greenozon, line 80: comment is misleading

whoami730: umm so this variable is being used for two different purposes - corresponds to `locals` before 3.11 and `localsplusnames` after 3.11. More like `locals` has been repurposed in cpython to include non local and global variables as well.

whoami730:
@zrax Hoping you can take a look at this one too. Thanks again!

Issue #573, opened 2025-08-29
When is Python3.13 supported?


1111qqjwbbw:
Current software does not support mainly getting stuck in an opcode.Unsupported opcode: MAKE_FUNCTION (122)Is an unsupported opcode.

Issue #569, opened 2025-07-31
Unsupported opcode: MAKE_FUNCTION (122)


samthemanekshahw:
Unsupported opcode: MAKE_FUNCTION (122) for a .pyc file built on Python 3.13 , how to decmpile it?

greenozon:
@samthemanekshahw   well!
there is a draft PR - https://github.com/zrax/pycdc/pull/539

you could give it a try on your sample!

Issue #568, opened 2025-07-30
Unsupported opcode: MAKE_FUNCTION (122)


zn123:
Unsupported opcode: MAKE_FUNCTION (122)

[main.zip](https://github.com/user-attachments/files/21505025/main.zip)

Issue #542, opened 2025-03-19
python 3.13 Unsupported opcode: MAKE_FUNCTION (122)


sh1man:

zhuxiaodics6:
same

ClickCyber:
anything?

greenozon:
FTR
https://docs.python.org/3/library/dis.html#opcode-MAKE_FUNCTION

PR #579, opened 2025-09-03
MAKE_FUNCTION & SET_FUNCTION_ATTRIBUTE_A for 3.13 

CarrBen:
# DRAFT again...
I'm setting this one up as another Draft, as this builds upon the changes in #539 & #538 

I've mainly done this because these things naturally build on top of each other as the Opcode evolve when moving forward through the Python versions. For example, the handling of Function arguments within SET_FUNCTION_ATTRIBUTE_A is the same as MAKE_FUNCTION in Python 3.6-3.12.

# Summary
* Handle the new MAKE_FUNCTION Opcode with the MAKE_FUNCTION_A path.
* Change MAKE_FUNCTION_A to just push a basic ASTFunction for 3.13, and not do anything with the Operand.
* Add support for the new SET_FUNCTION_ATTRIBUTE_A Opcode, using code from the previous PR that adds support for 3.6-3.12, to add args & kwargs to the basic ASTFunction pushed by MAKE_FUNCTION_A for 3.13.
* Update CALL to use the [new Stack Order for 3.13](https://docs.python.org/3/library/dis.html#opcode-CALL).
* Update LOAD_ATTR with [undocumented](https://github.com/python/cpython/issues/138402) 3.13 changes.
* Update the printing of source for NODE_STORE where the destination is a NODE_NAME to hide [new 3.13 dunder attributes](https://docs.python.org/3/whatsnew/3.13.html#:~:text=__static_attributes__);
  * \_\_static_attributes\_\_
  * \_\_firstlineno\_\_

## Tests
* Add 3.13 Test case for MAKE_FUNCTION & SET_FUNCTION_ATTRIBUTE.
  * Fails without this PR as MAKE_FUNCTION is unsupported.
  * Based on the test case for the previous MAKE_FUNCTION branches that have some additional args & kwargs functions.
* Add load_method Test case that fails due to a LOAD_ATTR error in 3.13.
  * Once MAKE_FUNCTION is in place, this test case has lots of None's, where NULLs end up in the wrong place in the method calls at the bottom of the file due to the new CALL Stack Order.

# Related Issues
#573 
#569 
#568 
#542 
#498


CarrBen:
I started looking at CALL_KW and found a bug in my LOAD_ATTR implementation, I put too much inside the if(operand).
LOAD_GLOBAL also needs a similar change to put things on the stack in the correct order

whoami730:
Thanks a lot for your contribution @CarrBen .

Unfortunately there seems to be a bug with `LOAD_GLOBAL` or similar for the below set of py and corresponding pyc file.
[failing_sample.zip](https://github.com/user-attachments/files/22144032/failing_sample.zip)

CarrBen:
Thanks for taking a look @whoami730, I had spotted an issue with LOAD_ATTR last night, and I'd completely missed updating LOAD_GLOBAL to work with 3.13.

I've pushed a further commit with updates for both Opcodes and it passes your test case on my end.

PR #590, opened 2025-10-21
Create c-cpp.yml

dmero2724-jpg:


zrax:
There are already CI builds configured here.

Issue #595, opened 2025-12-28
Unsupported opcode: MAKE_FUNCTION (0x7a) when decompiling Python 3.13 bytecode (practical example)


youlipeierges:
**Issue Description**
When using pycdc to decompile a `.pyc` file compiled with Python 3.13, the tool fails with the error: `Unsupported opcode: MAKE_FUNCTION (122)`. The error occurs when processing a practical application file (filename redacted for privacy).
﻿
**Environment**
- **Operating System**: Linux (CentOS Stream 9 x86_64)
- **Python Version**: 3.13 (version used to compile the `.pyc` file)
- **pycdc Version**: The version you are using (e.g., latest build from master branch)

Ennyolar96:
> **Issue Description** When using pycdc to decompile a `.pyc` file compiled with Python 3.13, the tool fails with the error: `Unsupported opcode: MAKE_FUNCTION (122)`. The error occurs when processing a practical application file (filename redacted for privacy). ﻿ **Environment**
> 
> * **Operating System**: Linux (CentOS Stream 9 x86_64)
> * **Python Version**: 3.13 (version used to compile the `.pyc` file)
> * **pycdc Version**: The version you are using (e.g., latest build from master branch)

Are you able to decompile the code? I was stuck on this

Issue #587, opened 2025-10-13
unsupported opcode:call_kw(247) from 3.13


morpheus315:
# Source Generated with Decompyle++
# File: 0401.pyc (Python 3.13)

Unsupported opcode: CALL_KW (247)
import sys
# WARNING: Decompyle incomplete
this is the pyc file:
[0401.txt](https://github.com/user-attachments/files/22887137/0401.txt)

PR #581, opened 2025-09-06
Version-aware MAKE_FUNCTION_A and MAKE_CLOSURE_A support

plfj:
## Summary
This PR replaces the existing counts-only implementation of `MAKE_FUNCTION_A` and `MAKE_CLOSURE_A` with a version-aware handler that supports all Python 3.x releases up to 3.14+.  

## Previous behavior
- Only supported operand format with **defCount** and **kwDefCount**.  
- Did not handle annotations, closure cells, or flag-based operand encodings.  
- Effective for Python ≤ 3.3 (and partially 3.4), but failed on later versions.  

## New behavior
- **Python ≤ 3.3**  
  - Retains counts-based handling (positional + keyword defaults).  
- **Python 3.4–3.5**  
  - Adds annotation support and adjusts argument default order (kw before positional).  
- **Python 3.6–3.9**  
  - Switches to **flag-based operand encoding**.  
  - Handles closure cells, annotation dicts, keyword defaults, and positional defaults.  
- **Python 3.10–3.13**  
  - Supports **PEP 563 deferred evaluation of annotations** (annotation may be dict or string).  
- **Python ≥ 3.14**  
  - Adds support for new operand flag (`0x10`), ensuring forward compatibility.  

## Outcome
- Functions are now reconstructed consistently across all supported Python versions.  
- AST generation correctly accounts for defaults, keyword defaults, annotations, and closure cells.  
- Backward compatibility with earlier operand encoding is preserved.  
-

Issue #547, opened 2025-04-29
Unhandled Opcodes


Aasyaco:
- *JUMP_NO_INTERRUPT*  
- *LOAD_SUPER_METHOD*  
- *LOAD_ZERO_SUPER_ATTR*  
- *LOAD_ZERO_SUPER_METHOD*  
- *SETUP_CLEANUP*  
- *STORE_FAST_MAYBE_NULL*

greenozon:
you should specify what Python version 
cause diff versions might handle same opcode differently

Aasyaco:
The opcodes of 3.13

PR #601, opened 2026-04-22
Python 3.13: CALL self-slot fix, CALL_KW, and assorted missing opcodes

sfz001:
## Summary

Adds the Python 3.13 call-path opcodes that pycdc was still missing, and fixes a latent bug in `CALL_A` where the direction of the `NULL` self-slot surrounding the callable was hard-coded.

### The call self-slot direction bug

In 3.11+, every `CALL` expects both a callable and a `NULL` self-slot on the stack. Which one is deeper depends on how the callable was loaded:

- `LOAD_GLOBAL` with the `+NULL` flag pushes `[NULL, callable]` (NULL deeper);
- `LOAD_ATTR` followed by a separate `PUSH_NULL` pushes `[callable, NULL]` (NULL on top).

The second form is the common pattern for non-method calls in 3.13 (e.g. `json.dumps({}, ensure_ascii=False)`). The existing handler only looked for the first layout, so those calls decompiled as `None(...)`. We now check for both arrangements.

### New opcodes handled

| Opcode | Behaviour | Related issue |
|---|---|---|
| `MAKE_FUNCTION` (no operand, 3.13) | Pop code object, push empty `ASTFunction` | #542, #569, #595 |
| `SET_FUNCTION_ATTRIBUTE` | Keep function on top, drop the attribute value | |
| `CALL_KW` | KW-names tuple at TOS, operand = total arg count | #587 |
| `TO_BOOL` | no-op for decompile | #540 |
| `MAKE_CELL`, `COPY_FREE_VARS` | prologue ops, no stack effect | |
| `LOAD_FAST_CHECK`, `LOAD_FAST_AND_CLEAR` | behave like `LOAD_FAST` | |
| `STORE_FAST_STORE_FAST`, `STORE_FAST_LOAD_FAST` | split into component ops | |
| `POP_JUMP_IF_NONE`, `POP_JUMP_IF_NOT_NONE` | reuse `POP_JUMP_IF_FALSE/TRUE` pipeline | |

The `SET_FUNCTION_ATTRIBUTE` handler currently discards the attribute value (defaults/kwdefaults/annotations/closure cells). That is intentionally minimal — PRs #579 and #581 have more complete treatments. Until one of those lands, this at least lets functions decompile instead of silently crashing.

## Test

Added `tests/input/call_kw_3_13.py`, compiled as `.3.13.pyc`, covering both the `CALL_KW` and `LOAD_ATTR + PUSH_NULL + CALL_KW` patterns:

```python
import json

def greet(name, greeting='hello'):
    return greeting + ', ' + name

print(greet('world', greeting='hi'))
print(json.dumps({}, ensure_ascii=False))
```

Before this patch the last line decompiled to `None({}, ensure_ascii=False)`.

## What this does not fix

Still missing from 3.13 decompilation (filed in #547 and elsewhere):

- `FORMAT_SIMPLE` / `CONVERT_VALUE` (f-strings)
- `CALL_INTRINSIC_1` / `CALL_INTRINSIC_2` (most variants)
- `CALL_FUNCTION_EX`
- `DICT_MERGE` / `DICT_UPDATE` / `MAP_ADD`
- PEP 695 class-parameter nodes (`NODE_TYPEPARAM`, seen as `Unsupported Node type: 27`)

I have ad-hoc stubs for a few of those in a working branch, but they're hacks; better to leave them for a follow-up once someone implements them properly.


ASTree.cpp
```diff
@@ -1647,6 +1709,39 @@ PycRef<ASTNode> BuildFromCode(PycRef<PycCode> code, PycModule* mod)
...
+            break;
+        case Pyc::SET_FUNCTION_ATTRIBUTE_A:
+            {
+                // Python 3.13+: TOS is the function produced by
+                // MAKE_FUNCTION, and the value below it is the
+                // attribute value (defaults tuple, kw-defaults dict,
+                // annotations, or closure cells -- selected by
+                // operand). Keep the function on top; discard the
+                // attribute value for now since pycdc does not yet
+                // render these extras.
```
zrax, line 1733: Suggestion: Might be worth adding an xfail case for these attributes if possible, so it's easier to see what's missing...

ASTree.cpp
```diff
@@ -1677,6 +1772,66 @@ PycRef<ASTNode> BuildFromCode(PycRef<PycCode> code, PycModule* mod)
...
+            break;
+        case Pyc::COPY_FREE_VARS_A:
+            // Python 3.11+: prologue that copies free variables from
+            // the function object into the frame. pycdc resolves free
+            // variable names via PycCode::getCellVar(), so this is a
+            // no-op during decompilation.
+            break;
+        case Pyc::LOAD_FAST_CHECK_A:
+            // Python 3.12+: like LOAD_FAST but raises if unbound. For
+            // decompilation purposes they are interchangeable.
```
zrax, line 1794: Probably worth merging these cases rather than duplicating their content.  The comments can be added under the case labels, with a fallthrough attribute if the compiler requires it.

PR #608, opened 2026-06-08
Python 3.11 decompilation fixes (for/while/try/with/comprehensions/comparisons)

goromachine:
## Summary

This PR adds comprehensive Python 3.11 decompilation support. Python 3.11
introduced several bytecode changes that break the existing decompiler logic.
All changes are backward-compatible (guarded by version checks where needed).

Validated against 239 real-world Python 3.11 `.pyc` files from a production
codebase (all 239 produce output that compiles as valid Python 3.11 via
`py_compile`) and 95 Python 3.11 stdlib modules (45/95 fully clean output).

> **Note on process**: these fixes were developed with significant assistance
> from Claude (Anthropic), reflected in the `Co-Authored-By` lines. The fixes
> are individually tested and verified against a regression harness before each
> commit.

---

### New opcodes handled
- `MAKE_CELL`, `COPY_FREE_VARS`, `LOAD_ASSERTION_ERROR`
- `DICT_MERGE`/`DICT_UPDATE`, `MAP_ADD`, `LIST_TO_TUPLE`
- `RETURN_GENERATOR` (generators/coroutines)
- `POP_JUMP_FORWARD/BACKWARD_IF_TRUE/FALSE/NONE/NOT_NONE`

### Control flow reconstruction

**`for` loops** — 3.11 removes `POP_BLOCK`; the loop now closes via
`JUMP_BACKWARD` whose operand is relative (×2, `pos − operand`). Previously
the loop never closed because the code compared the raw offset against the
absolute start position; code after the loop was silently absorbed into the body.
`JUMP_BACKWARD` before the loop end is emitted as `continue`.

**`while` loops** — 3.11 bottom-test optimization: `POP_JUMP_FORWARD_IF_FALSE end; body; POP_JUMP_BACKWARD_IF_TRUE start`. New pre-pass `ScanWhileLoops` detects the guard+back-edge pair and reconstructs the `while` block correctly (was emitting two unrelated `if` statements).

**`with` statement** — 3.11 mid-block shape with `BEFORE_WITH` + exception-table cleanup (no `SETUP_WITH` or `POP_BLOCK`).

**`try/finally`** — 3.11 exception-table based; no `SETUP_FINALLY`.

**`try/except` with nested `try`** — the exception table splits a try region into multiple entries sharing one handler. Previously a new spurious `try` was opened for each entry producing cascading nesting. Now detects continuation of an already-open `try`.

**Chained comparisons** (`a == b == c`) — compiled with `COPY`/`SWAP` plus a `POP_TOP; JUMP_FORWARD` trampoline. New pre-pass `ScanChainedCompare` resolves the trampoline and emits `a == b and b == c`.

### Exception handling

**`except ... as e`** — 3.11 emits `PUSH_EXC_INFO; CHECK_EXC_MATCH; STORE_FAST e` plus cleanup (`LOAD_CONST None; STORE_FAST e; DELETE_FAST e`). The cleanup epilogue is now suppressed in output.

**`raise X from Y`** — rendered correctly for Python 3 (was Python 2 `raise X, Y`).

### Expressions and comprehensions

**`SWAP_A`** — treated as a real stack swap (needed for chained comparisons and starred unpack; was a no-op).

**List/dict/set/generator comprehensions** — inline the implicit code object call, substitute the `.0` argument, support multiple nested filters.

**Class kwargs** (`metaclass=` and other keyword arguments in class definitions) — reconstructed from 3.11 `KW_NAMES`/`CALL` bytecode.

### Correctness fixes

- **Return-in-if**: only skip the instruction after `return` inside an if/else block when it is a real unconditional jump — not arbitrary code.
- **`continue` in `else`**: emit `continue` for a backward jump out of an `else` into the enclosing loop.
- **Function signature**: `*args` placed before keyword-only parameters.
- **Decorator reconstruction**: 3.11 no-`PUSH_NULL` form.
- **Class artifacts**: strip `__classcell__` / `return __class__` epilogue.
- **Module-level returns**: strip spurious `return None` at module level (all nested blocks).
- **EOF guard**: don't read past end of bytecode buffer on empty bodies.
- **`except` boundary**: handle clause that raises / clause at function end.


Dark52s:
bro your job its so good, thanks for the parch

